PR #1, opened 2023-10-16
:fnplead:

McAwesome123:


bitmaps/fnplead.png
pierce-smith1, line 1: It might be better for the dress to be based off horn and eye color rather than horn and horn shadow color, since the former is more likely to contrast more sharply (like the actual dress) and generally be more interesting.
Maybe the dress is also just hard coded to blue and yellow like the real thing, but that sounds boring.

spritecontrol.cpp
```diff
@@ -24,7 +24,7 @@ std::vector<Sprite *> SpriteGenerator::make(unsigned int n) const {
 
 	for (float y = -1.2f; y < 1.2f; y += 1.0f / sqrt(cfg.at(SpriteCount))) {
 		for (float x = -1.2f; x < 1.2f; x += 1.0f / sqrt(cfg.at(SpriteCount))) {
-			if (Noise::random() < 0.001f) {
```
pierce-smith1: I'd prefer if this didn't change at all, but it's such a small difference I might not care. I'll figure out when I test it later.

pierce-smith1 (commented):
Hell yeah!

bitmaps/fnplead.png
McAwesome123: Yeah, I thought about using another color, but I was only looking at the ones I hadn't already used.

spritecontrol.cpp
McAwesome123: I only increased it mainly because there's more of them but it shouldn't matter so I've reverted it.

McAwesome123:
I've changed the dress colors and reverted the impostor chance change. Let me know if you still have concerns.

pierce-smith1 (commented):
I'll stamp a ✅ on this when I have time to compile and test it.

pierce-smith1 (approved):

PR #5, opened 2023-10-23
:lotusyoy:

McAwesome123:


sprite.cpp
```diff
@@ -150,3 +150,12 @@ BitmapName Impostor::random_bitmap() {
 	static std::vector<BitmapName> impostors = { cvjoy, nx, vx, lkmoyai, fn, fnplead };
 	return impostors[(int) (Noise::random() * impostors.size())];
 }
+
+// <Insert funny Pierce comment>
+Yoy::Yoy(PaletteName palette, const Point& home)
```
pierce-smith1: I'm nitpicking, but let's be consistent with styling. In this house, the ampersand goes on the right!

sprite.h
```diff
@@ -65,3 +65,11 @@ class Impostor : public Sprite {
 protected:
 	static BitmapName random_bitmap();
 };
+
+class Yoy : public Sprite {
+public:
+	Yoy(PaletteName palette, const Point& home);
```
pierce-smith1: Styling: to the right y'all!

resources.h
```diff
@@ -42,7 +46,11 @@ const static std::map<BitmapName, int> BITMAP_RESOURCE_IDS({
 	{ lkxd, IDB_BITMAP15 },
 	{ nx, IDB_BITMAP16 },
 	{ vx, IDB_BITMAP17 },
-});
+	{ lkyoy, IDB_BITMAP18 },
+	{ lkyoyapprove, IDB_BITMAP19 },
+	{ fnyoy, IDB_BITMAP20 },
+	{ cvyoy, IDB_BITMAP21 },
+	});
```
pierce-smith1: Styling: make sure that closing brace is all the way to the left.

resourcew.h
```diff
@@ -20,6 +20,10 @@
 #define IDB_BITMAP15                    152
 #define IDB_BITMAP16                    153
 #define IDB_BITMAP17                    154
+#define IDB_BITMAP18                    155
+#define IDB_BITMAP19                    156
+#define IDB_BITMAP20                    157
+#define IDB_BITMAP21                    158
```
pierce-smith1, line 26: Very pleased you were able to figure out this whole song and dance, nicely done.

sprite.h
```diff
@@ -65,3 +65,11 @@ class Impostor : public Sprite {
 protected:
 	static BitmapName random_bitmap();
 };
+
+class Yoy : public Sprite {
```
pierce-smith1: Architecture: I'm not a huge fan of making an entire new class for yoys here. The `Sprite` class is a base for customizing the **behavior** of the characters - i.e., the difference between a `Yonker` and an `Impostor` is one of behavior (Yonkers change texture and jitter around, Impostors do not.) The `Yoy` here is behaviorally identical to an `Impostor`.

That said, my suggestion is to put the yoys into the impostors. I see there's an intent here to make the yoys and impostors evenly distributed, so that there's a 50/50 chance of either - to keep that behavior I might prefer to modify the `Impostor` class to draw from two separate lists in `random_bitmap()`, with a 50/50 chance of selecting either list. Something like:

```c++
BitmapName Impostor::random_bitmap() {
	static std::vector<BitmapName> impostors = { cvjoy, nx, vx, lkmoyai, fn, fnplead };
	static std::vector<BitmapName> yoy = { lkyoy, lkyoyapprove, fnyoy, cvyoy };

        if (Noise::random() < 0.5f) {
	    return impostors[(int) (Noise::random() * impostors.size())];
        } else {
            return yoys[(int) (Noise::random() * yoys.size())];
        }
}
```

pierce-smith1 (changes requested):
Awesome work! A few concerns I want to at least talk about before this gets merged.

source_bitmaps/lkyoy.png
pierce-smith1, line 1: Didn't catch this in my original review, but if you take a look at the existing lkunamused and lkthumbsup bitmaps, you'll see they've been smoothed over a bit to remove the scale and horn details. That was an intentional stylistic choice and I think it should be consistent with the yoys.

resources.h
McAwesome123: this should be fixed

source_bitmaps/lkyoy.png
McAwesome123: should be changed

sprite.cpp
McAwesome123: this section has been removed

sprite.h
McAwesome123: Never thought about doing it like this. And imo it's way better. I've changed it.

sprite.h
McAwesome123: :lkanim:

McAwesome123:
I've changed the yoys and merged the two classes.
Lemme know if there's anything else I've missed.

pierce-smith1 (approved):
Good 'nuff.

PR #17, opened 2023-10-30
Allow patterns to be fixed

### LINKED ISSUE #8: Allow patterns to be fixed
Right now, the screensaver inevitably cycles through each of the available patterns, and all you can do is increase the delay between changes. It's been requested before that there be a way to fix a pattern so that it never changes.

pierce-smith1:
#8. Adds a config option that allows you to say a single pattern should play indefinitely.

@McAwesome123 - This will conflict with things you're working on, but that's okay, we will probably be able to resolve those easily.


resourcew.h
```diff
@@ -1,49 +1,51 @@
...
-// 
-#ifdef APSTUDIO_INVOKED
-#ifndef APSTUDIO_READONLY_SYMBOLS
-#define _APS_NEXT_RESOURCE_VALUE        159
-#define _APS_NEXT_COMMAND_VALUE         40001
-#define _APS_NEXT_CONTROL_VALUE         1019
-#define _APS_NEXT_SYMED_VALUE           101
-#endif
-#endif
+//{{NO_DEPENDENCIES}}
```
pierce-smith1, line 1: Not sure why git is saying these files are almost entirely different, but okay.

McAwesome123 (approved):

PR #23, opened 2023-11-01
Impostors and Palettes

### LINKED ISSUE #7: Separate palettes into banks which can be independently toggled
It feels kind of weird to have both canon yokers and non-canon yokers both appearing at once. I think it would be cool to separate all of the palettes we have into groups, which can independently be toggled on/off and maybe even made more/less frequent. That way, if someone like me prefers to only have the non-canon yokers, they can. These kind of structural changes will also help us keep palettes organized and less chaotic if and when we make more of them.

### LINKED ISSUE #9: Make impostor frequency a slider
Since the impostors do look pretty neat and there are quite a few of them now, it'd be cool to customize how frequently they show up. I don't have any problems with allowing someone to go anywhere from 0% to 100% impostors, but we need to make sure the UI still "prefers" to keep them rare.

### LINKED ISSUE #12: Make the build artifact called yok.scr
VS builds our project into an executable called `yokscr.exe`, but we really want it to be called `yok.scr`. I have no idea if this is even possible, but I figure it must be.

McAwesome123:


McAwesome123:
Need to fix the crash from the color count being set too high when changing the palettes option to something other than all.

configdialog.cpp
```diff
@@ -45,8 +50,15 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 		}
 		case IDC_YONK_PATTERN: {
 			if (HIWORD(wparam) == CBN_SELENDOK) {
-				return combobox_changed((HWND) lparam);
+				return combobox_changed((HWND) lparam, IDC_YONK_PATTERN);
```
pierce-smith1, line 53: You might be surprised to hear but I have no problem with how this is being done.

resources.h
```diff
@@ -73,42 +100,129 @@ enum PaletteName {
...
+		PaletteName::ripple,
+		PaletteName::stonehenge,
+		PaletteName::tulips,
+		PaletteName::vortecspace,
+		PaletteName::wind,
+		PaletteName::windowsxp,
+		PaletteName::yette,
+		PaletteName::zehal,
+	}},
+	{ PaletteGroup::All, {	// :lkfear:
```
pierce-smith1: Let's do whatever we can to avoid repeating ourselves here since this is just asking for trouble. Here's an idea: since these palette names are just the numbers from 0 to 47 or whatever, let's just generate it at run time with a function:

```c++
inline std::vector<PaletteName> make_all_palettes() {
    std::vector<PaletteName> all_palettes((size_t) PaletteName::_PALETTE_COUNT);
    std::generate(all_palettes.begin(), all_palettes.end(), [n = 0]() mutable { return (PaletteName) n++; });
    return all_palettes;
}

// ...
{ PaletteGroup::All, make_all_palettes() }
```

This is some of the weirdest fucking c++ I've ever seen but it does seem to work. I stole it from [here](https://en.cppreference.com/w/cpp/algorithm/generate).

spritecontrol.cpp
```diff
@@ -9,12 +9,13 @@ SpriteGenerator::SpriteGenerator() {
 	using namespace std::chrono;
 	std::srand(duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count());
 
-	std::vector<int> bag_of_palettes(_PALETTE_COUNT);
-	std::iota(bag_of_palettes.begin(), bag_of_palettes.end(), 0);
+	//std::vector<int> bag_of_palettes(static_cast<int>(PaletteName::_PALETTE_COUNT));
```
pierce-smith1: Be sure to clean these comments up.

configdialog.cpp
```diff
@@ -92,8 +111,23 @@ void ConfigDialog::refresh() {
 		SendMessage(slider, TBM_SETPOS, TRUE, encodef(m_current_config.at(entry.first)));
 	}
 
+	if (m_current_config.at(YonkPattern) < 0 || m_current_config.at(YonkPattern) >= _PATTERN_COUNT)
```
pierce-smith1: Something about all of this is hurting my 3 am brain and I don't know what to make of it. I'll come back to this.

pierce-smith1 (changes requested):

spritecontrol.cpp
```diff
@@ -2,19 +2,23 @@
...
+	//std::iota(bag_of_palettes.begin(), bag_of_palettes.end(), 0);
+	std::vector<PaletteName> bag_of_palettes = PALETTES_BY_GROUP.at((PaletteGroup) (cfg.at(YonkPalette)));
 
-	for (int i = 0; i < round(cfg.at(MaxColors)); i++) {
+	float maxColors = cfg.at(MaxColors) * (bag_of_palettes.size() / config_ranges.at(MaxColors).second);
```
pierce-smith1: We really want `MaxColors` to be a value between 0.0 and 1.0, but we can't just do that because people's existing configs will be outside of that range. So I guess this is fine for now.

pierce-smith1: However - styling!! `snake_case` for locals please. I'm pointing this out because VS won't enforce it, even with the new .editorconfig.

spritecontrol.cpp
```diff
@@ -2,19 +2,23 @@
...
+
+	for (int i = 0; i < round(maxColors); i++) {
```
pierce-smith1: I'm scared of this. If the result of `round(maxColors)` is greater than `bag_of_palettes.size()`, the program will try to erase from an empty list and invoke undefined behavior. We cannot mathematically reason that this will never be the case because of floating-point imprecision. 

Since it's so important for this to not exceed `bag_of_palettes.size()`, I think we should do something like this, even though it is pretty clunky:

```c++
if (round(maxColors) > bag_of_palettes.size()) {
    maxColors = bag_of_palettes.size();
}
```

spritecontrol.cpp
```diff
@@ -24,7 +28,7 @@ std::vector<Sprite *> SpriteGenerator::make(unsigned int n) const {
 
 	for (float y = -1.2f; y < 1.2f; y += 1.0f / sqrt(cfg.at(SpriteCount))) {
 		for (float x = -1.2f; x < 1.2f; x += 1.0f / sqrt(cfg.at(SpriteCount))) {
-			if (Noise::random() < 0.002f) {
+			if (Noise::random() < pow(cfg.at(ImpostorChance), 3)) {
```
pierce-smith1, line 30: I may have said it already, but the cubic is pretty clever here.

configdialog.cpp
pierce-smith1: OK, here are some thoughts. First, I don't think we should show a message box in these cases, we should just silently fix the value and move on. I have a lot to say about this but really it just boils down to "I don't want to annoy users with a message box over something this unimportant."

Second, it feels like this validation + repair stuff really wants to be a separate system in the code. It might be as simple as just having a function `is_config_value_in_range` that we can call rather than doing these if checks manually. I want to think about this some more though, because I'm not 100% sure how we should design it, so I'll make another issue for it and we can address it later.

configdialog.cpp
```diff
@@ -92,8 +111,23 @@ void ConfigDialog::refresh() {
...
+		MessageBox(NULL, (LPCWSTR)L"Yonk Pattern was set to an out of range option!\nIt will be reset to default.", (LPCWSTR)L"Yok.scr", MB_ICONWARNING | MB_OK);
+		m_current_config.at(YonkPattern) = 0;
+	}
 	HWND pattern_box = GetDlgItem(m_dialog, IDC_YONK_PATTERN);
 	ComboBox_SelectString(pattern_box, -1, pattern_strings.at(
-		(SpritePattern) m_current_config.at(YonkPattern)).c_str()
+		(SpritePattern)m_current_config.at(YonkPattern)).c_str()
+	);
+
+	if (m_current_config.at(YonkPalette) < 0 || m_current_config.at(YonkPalette) >= static_cast<int>(PaletteGroup::_PALETTE_OPTION_COUNT))
```
pierce-smith1: Why does there need to be a `static_cast` here, but not on the previous if? And as a style point (that VS is not set up to enforce), we should use C style casts instead of `static_cast`.

Resource.rc
pierce-smith1, line 1: image

Two comments on the dialog: 
* The label "Palette" should maybe be pluralized into "Palettes," since we don't want people getting confused and thinking they're selecting one palette for all the yokers.
* The whole "Family" box a just a touch too tall, with a little bit too much empty space on the bottom. It could be shrunk just a tiny bit.

pierce-smith1 (changes requested):
I think this got everything out of my system. I'm going to wait until this PR gets merged before merging any of my other ones, because I want to impending apocalyptic merge conflict shitstorm to fall on me, not you.

configdialog.cpp
McAwesome123: Wasn't aware that a C-style cast works here. And it's needed because the compiler doesn't like comparing an int to an enum class member.

McAwesome123:
6 am programming goes brrrr

pierce-smith1 (approved):
It's release time motherfucker.

Issue #25, opened 2023-11-03
Make the version user-visible


pierce-smith1:
We should let people know what version of the screensaver they're running, probably by putting something in the config dialog.

PR #31, opened 2024-01-06
Add version number to config dialog

pierce-smith1:
#25 

We're gonna have to remember to update the version string in `yokscr.h` every time we make a release. I don't know how the fuck that's going to work. There's probably some super fancy automatic github shit we could set up but it probably isn't worth it.


resourcew.h
```diff
@@ -47,7 +48,7 @@
 #ifndef APSTUDIO_READONLY_SYMBOLS
 #define _APS_NEXT_RESOURCE_VALUE        159
 #define _APS_NEXT_COMMAND_VALUE         40001
-#define _APS_NEXT_CONTROL_VALUE         1020
+#define _APS_NEXT_CONTROL_VALUE         1021
```
pierce-smith1, line 51: The Win32 API continues to entertain me with its naivety.

_SURELY there's nothing that can go wrong with making every new control have an ID that's just one larger than the previous, right? That won't possibly mean that two people working in different branches will end up with controls with the same ID and have a nightmare on their hands, right?_

Different times.

McAwesome123: Speaking of this, I feel the need to mention that there's are `IDC_CHECK1` and `IDC_COMBO1` defines that have the same value as something else and are completely unused as far as I can tell.

McAwesome123 (approved):
Seems good, just don't forget to change the version number for the next release :)

resourcew.h
pierce-smith1: We should probably get rid of those, but they're most likely just artifacts from visual studio and aren't causing any trouble.

Issue #26, opened 2024-01-06
Random palette generation


pierce-smith1:
This has been suggested a few times, and it probably wouldn't be too hard to do. We can start with a basic "totally random" generator, but it might be fun to go deeper and actually try to _mostly_ generate sensible yokin colors using some extra artistic smarts.

This could entail _every_ yoker being a random color, or just generating a random set of palettes. The latter is more consistent with existing options, but the former sounds like a lot of fun. Maybe offer both?

PR #32, opened 2024-01-07
Random palette generation

pierce-smith1:
#26 

This PR adds an option, "I'm Feeling Lucky", to the palette dropdown, which will randomly generate a group of palettes rather than using any of the predefined ones. It also restructures and moves the existing palette-related code and introduces a `common.h` file for generic utility classes and functions.

The random generation is very finely tuned to be both interesting and easy on the eyes.

### DIFF
--- a/graphics.cpp
+++ b/graphics.cpp
@@ -1,17 +1,10 @@
 #include "graphics.h"
 #include "resources.h"
 
-// Here, textures are born not by loading from disk...
-Palette::Palette(const std::initializer_list<Color> &i_list) {
-	std::copy(i_list.begin(), i_list.end(), begin());
-}
-
-// But by our own hands, with ingredients.
 Bitmap::Bitmap(const std::initializer_list<GLubyte> &i_list) {
 	std::copy(i_list.begin(), i_list.end(), begin());
 }
 
-// Lazily baked, not wasting our time...
 const Texture *Texture::get(const Palette &palette, const Bitmap &bitmap) {
 	auto ids = std::make_pair(palette.id(), bitmap.id());
 	auto result = texture_cache.find(ids);
@@ -23,7 +16,6 @@ const Texture *Texture::get(const Palette &palette, const Bitmap &bitmap) {
 	return texture_cache.at(ids);
 }
 
-// And make calling convenient, 'fore I wish any crimes.
 const Texture *Texture::of(PaletteName palette_name, BitmapName bitmap_name) {
 	return get(PALETTES.at(palette_name), *load_bitmap(bitmap_name));
 }
--- a/graphics.cpp
+++ b/graphics.cpp
@@ -20,6 +20,10 @@ const Texture *Texture::of(PaletteName palette_name, BitmapName bitmap_name) {
 	return get(PALETTES.at(palette_name), *load_bitmap(bitmap_name));
 }
 
+const Texture *Texture::of(const Palette *palette, BitmapName bitmap_name) {
+	return get(*palette, *load_bitmap(bitmap_name));
+}
+
 Bitmap::Bitmap(const GLubyte *data) {
 	std::copy(data, data + size(), begin());
 }
--- a/palettes.cpp
+++ b/palettes.cpp
@@ -1,5 +1,6 @@
 #include <map>
 #include <algorithm>
+#include <vector>
 
 #include "palettes.h"
 #include "common.h"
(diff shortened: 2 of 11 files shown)


graphics.cpp
```diff
@@ -1,17 +1,10 @@
 #include "graphics.h"
 #include "resources.h"
 
-// Here, textures are born not by loading from disk...
```
pierce-smith1, line 4: Unfortunately I had to kill this poem because part of it had to get moved to another file.

McAwesome123 (changes requested):
Random palettes are nice, but the method of generating them seems a bit flawed.
For one, they are more or less slight variations of a randomly picked colour, meaning that if said colour is desaturated, most of them end up looking pretty dull.
Additionally, the population diversity setting ends up not doing much as it just adds more palettes without changing how they are generated, so they end up being mostly the same colours.

pierce-smith1:
Well this exposes an interesting debate in my head. And some flaws with this PR.

tl;dr, it seems reasonable to:

* Get rid of color biasing so all hues have an equal chance of appearing
* Add some logic to make colors more saturated on average

> they are more or less slight variations of a randomly picked colour

This was intentional, since if the colors are _completely_ random, then every single run of the random generator ends up looking about the same, just a mess of weird colors. I prefer it when every set of random palettes has a unique style that is noticeably different from the last one. 

But... this problem is already solved by the diversity slider. To achieve the effect I like, you don't even need to tune the generation, you just need to turn the max colors down. So takeaway, the color weighting seems unnecessary and should probably be ripped out. Every hue should have an equal chance of showing up.

Now here's the debate in my head. I would bet you good money that 90% of people using this screensaver right now, if they even bothered to open the settings page at all, have cranked their diversity all the way to max and are going to leave it there forever. I think a low max colors offers a much better experience, and I want to share that experience, but I also want to respect people's choices - if they asked for many colors, they should get many colors.

Ultimately I think the right call is the simplest one, just remove the biasing.

> if said colour is desaturated

These are yokins we're talking about, generally they shouldn't have desaturated colors, outside of their horns. I can see some ways to introduce logic that will greatly reduce the likelihood of desaturated colors and I want to do that.

Because of this it's quite tempting to generate colors in terms of HSL/HSV, but I would then have to manually convert them to RGB. It's not _that_ complicated of a process, but it is not trivial, and it would be awkward to have to shove into the codebase. Plus since all things are tradeoffs, there would probably be downsides to HSL/HSV that I can't forsee. All we want to do here is ensure colors are saturated, and you don't _need_ HSL/HSV for that, you just need to make sure that colors are not too evenly distributed among the channels.

McAwesome123:
I do understand wanting to bias the colours and to be honest, I was thinking something closer to keeping the colour bias, but loosening it based on the colour diversity setting.

I'm not too sure how this could be done using plain RGB. From my understanding of your code, you could, of course, just increase the "budget chunks", but that would come at the cost of fewer possible colour combinations. And probably other things.

HSV/HSL would probably be simpler with just needing to increase the hue range. It probably wouldn't even need to cover the entire range, just enough so that it's not entirely slightly different shades of red or green or yellow. This way, even the people who max out the diversity slider can get different colours every time.

Of course, the best option would probably be to add settings for the randomizer. But that would probably require reworking the UI to some extent, lest it become a confusing mess.

pierce-smith1:
> I was thinking something closer to keeping the colour bias, but loosening it based on the colour diversity setting.

I'm going to test how totally unbiased generation feels first. If it feels like it's still too messy even on low settings, I'll definitely consider it.

> I'm not too sure how this could be done using plain RGB... HSV/HSL would probably be simpler

It turns out that, if all you want to do is generate a random HSV color where S and V are both 100 and H is something random between 0 and 360, it's actually quite easy to do directly with RGB: you just choose one channel to be 0, one to be 255, and the last to be something random (source: spin the color wheel around in paint.net). From there the existing lightening/darkening functions can be applied for extra variety. So actually, my hubris says it should be pretty easy.

Time to try some of this out.

palettes.cpp
```diff
@@ -0,0 +1,186 @@
...
+		generated_palettes[rng_token] = new_random_palette();
+	}
+
+	return generated_palettes.at(rng_token);
+}
+
+Palette *RandomPalettes::new_random_palette() {
+	static int bias_intensity = (int) (30.0f / ((cfg.at(MaxColors) + 14.0f) / config_ranges.at(MaxColors).second));
+
+	static float red_bias = rand() % bias_intensity * (rand() % 2 ? -1.0f : 1.0f);
```
pierce-smith1, line 34: I'm aware that most of this math is really hard to read. Using the C++ standard library RNG may have helped the readability quite bit here but I HATE it, the interface and the performance are both absolute shit. So it's just gonna have to be write-only code for now.

McAwesome123 (approved):
The random palettes seem a lot better now.
It seems that with the diversity set to be more kindred, the palettes tend to be more similar to each other, although still with a few that stand out, while when set to be more vibrant, the palettes tend to be more diverse with overall fewer similarities.

PR #33, opened 2024-01-09
Proper registry value names

McAwesome123:
As I recently discovered, trying to manually adjust the config options from the registry is really difficult when all you have to go off of are numbers. So this PR should fix that.

They are now named the same way as in code. (Because that's the simplest.)

At the current moment, any "old" registry values should be automatically migrated when the screensaver runs for the first time.

### DIFF
--- a/config.cpp
+++ b/config.cpp
@@ -28,29 +28,29 @@ Registry::Registry()
 Config Registry::get_config() {
 	std::vector<std::pair<ConfigOption, float>> keys;
 
-	const std::wstring migrated_registry_name = L"MigratedRegistry";
-	bool migrated_registry = (bool) ((int) get(migrated_registry_name.c_str(), 0.0f));
+	const std::wstring is_registry_migrated_name = L"IsRegistryMigrated";
+	bool is_registry_migrated = get(is_registry_migrated_name, 0.0f) != 0.0f;
 
 	for (int i = _CONFIG_OPTIONS_START + 1; i < _CONFIG_OPTIONS_END; i++) {
 		ConfigOption opt = (ConfigOption) i;
-		if (migrated_registry) {
-			keys.push_back(std::make_pair(opt, get(config_names.at(opt).c_str(), cfg_defaults.at(opt))));
+		if (is_registry_migrated) {
+			keys.push_back(std::make_pair(opt, get(config_names.at(opt), cfg_defaults.at(opt))));
 		} else {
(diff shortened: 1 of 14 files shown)


McAwesome123:
me when i base the branch on another branch instead of master
(yes, master is technically still a branch but shhhh)

config.cpp
```diff
@@ -28,25 +28,37 @@ Registry::Registry()
 Config Registry::get_config() {
 	std::vector<std::pair<ConfigOption, float>> keys;
 
+	const std::wstring migrated_registry_name = L"MigratedRegistry";
+	bool migrated_registry = (bool) ((int) get(migrated_registry_name.c_str(), 0.0f));
```
pierce-smith1: This would be a lot clearer if it were called something like `is_registry_migrated`. Ditto for the name of the registry entry itself.

Casting to bool is also pretty yucky. What about:

```c++
bool is_registry_migrated = get(migrated_registry_name.c_str(), 0.0f) != 0.0f;
```

config.h
```diff
@@ -23,15 +23,32 @@ enum ConfigOption {
...
 class Registry {
 public:
 	Registry();
 
 	Config get_config();
-	float get(ConfigOption opt, float default_);
-	void write(ConfigOption opt, float value);
+	float get(LPCWSTR opt, float default_);
+	void write(LPCWSTR opt, float value);
+	void remove(LPCWSTR opt);
```
pierce-smith1: These three should just take in `const std::wstring &`s instead of `LPCWSTR`s. This will avoid having to `.c_str()` at every call site, and it also helps me forget the Win32 api exists.

config.h
```diff
@@ -23,15 +23,32 @@ enum ConfigOption {
 	_CONFIG_OPTIONS_END
 };
 
+const static std::map<int, std::wstring> config_names = {
+	{ YonkStepSize, L"YonkStepSize" },
```
pierce-smith1, line 27: Yuck. Gross. But I'm well aware if we want to do this there is no other way.

pierce-smith1 (approved):
Cool, this will help us out a lot I imagine.

pierce-smith1:
> me when i base the branch on another branch instead of master

Don't do it again!!!

config.h
McAwesome123: I don't disagree.

pierce-smith1 (approved):
Sure.

Issue #11, opened 2023-10-24
Make an option to not clear the screen


pierce-smith1:
A lot of classic screensavers have the option to let the screensaver play over a screenshot of your desktop rather than over a black background. It'd be cool for yokscr to support this as well. I have absolutely no idea how to do this.

**Hints**
We will probably just need to do exactly what I said, take a screenshot of the desktop, paint that screenshot as the background, then play the screensaver over it. [This may be what we need to do,](https://learn.microsoft.com/en-us/windows/win32/gdi/capturing-an-image) or at least set us on the right track.

PR #34, opened 2024-01-15
Add an option to not clear the screen

pierce-smith1:
#11 

This adds an option, "Play over the desktop", that renders the animation over a screenshot of the desktop rather than over a black screen. 

Some possible issues:

* When the screensaver starts, it seems like it sometimes might be flashing the screen black for a few milliseconds. I'd rather it not strobe the screen, but I don't know if there's anything I can do about it.


resources.cpp
```diff
@@ -19,22 +19,7 @@ Bitmap *load_bitmap(BitmapName name) {
...
-		for (size_t x = 0; x < BITMAP_WH; x++) {
-			int i_source = y * BITMAP_WH + x;
-			int i_target = (BITMAP_WH - y - 1) + BITMAP_WH * x;
-	
-			data[i_target] = ((GLubyte *) bitmap.bmBits)[i_source];
-		}
-	}	 
-	// ------------------------------------------------------------
-	//                      GET ROTATED
-	//                         Idiot
```
pierce-smith1, line 35: It turns out that, this whole time, I was actually drawing the sprites _sideways_ - which was the only reason I needed to rotate the bitmap here. I fix that in this PR so this rotating is no longer necessary.

McAwesome123: Huh, I was always wondering what you're rotating.

McAwesome123 (approved):
All looks good and I haven't noticed issues while testing it.

PR #35, opened 2024-01-17
Fix unintended side effects when increasing bravery

McAwesome123:
This makes the yokers not shake violently when increasing just the bravery setting (without increasing the shakiness). They will also not disappear for far too long when going off-screen at high bravery.

### DIFF
--- a/sprite.cpp
+++ b/sprite.cpp
@@ -8,7 +8,7 @@
 
 using std::get;
 
-Sprite::Sprite(const Texture *texture, const Point &home) 
+Sprite::Sprite(const Texture *texture, const Point &home)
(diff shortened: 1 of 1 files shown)


sprite.cpp
```diff
@@ -42,18 +42,21 @@ void Sprite::draw(Context &ctx) {
 }
 
 void Sprite::update(Context &ctx) {
-	auto wrap = [](float n, float min, float max) -> float {
-		if (n < min) {
-			return max - (min - n);
-		} else if (n > max) {
-			return min + (n - max);
+	auto wrap = [](float home, float total, float min, float max) -> float {
```
pierce-smith1, line 45: I'm not gonna hold up the PR over this but note to self, on some cleanup pass we should maybe change the name of this function and its parameters because the true intent is really confusing here.

pierce-smith1 (approved):

Issue #30, opened 2024-01-06
Consider dumping enums?


pierce-smith1:
There is a LOT of enum-to-integer casting going on in this codebase. At first enums seemed like the right call for things like config options and palette names, but we are doing an unexpected amount of math on these values, to the point where they maybe should just be integers. 

So, we could go through all the enums we have (that we do a lot of math on) and just turn them into constants:
```c++
enum class PaletteName {
    autumn,
    bliss,
    ...
}

// ... might become ...

const int PALETTE_AUTUMN = 0;
const int PALETTE_BLISS = 1;
...
```

The disadvantage here, of course, is that it's more error prone and it's way more annoying to define new variants. There is probably a more elegant way. But... boy, are these enum things getting irritating.

pierce-smith1:
A lot of enums have been removed, but those for patterns still exist.

PR #39, opened 2024-03-02
Replace enums for config options

pierce-smith1:
Part of #30, but will not necessarily close it.

This PR is an attempt at a new way of managing config options. Previously, options were defined as variants of an enum; now they are defined as static objects of a ConfigOption type. This has several advantages:

* Relevant information for each config option is now attached to the config option object itself, rather than being spread around the codebase in static maps. For instance now if you want the name for a config option, instead of calling `config_option_names[option]` or whatever (and having to go find the header where `config_option_names` is defined), you can do a much more readable `option.name`!
* There is now a standard list of config options. Previously, enumerating config options or sampling random ones would require math on the `enum` type, which is dumb. Now we can use standard c++ algorithms to do just about anything. This isn't super relevant on config options, but demonstrates how things will look if we use a similar strategy for other enums where this operation is more common. 

While this is all nice, I am still sort of on the fence if this effort is worth it.


config.h
```diff
@@ -4,39 +4,136 @@
...
+	};
+
+	inline const static ConfigOption Palette = {
+		.name = L"YonkPalette",
+		.legacy_id = 1013,
+		.default_ = 0.0f,
+		.dialog_control_id = IDC_YONK_PALETTE,
+	};
+
+	inline const static std::vector<ConfigOption> All = {
```
pierce-smith1: A disadvantage with this approach is that if you define a new config option, you must also remember to define it a second time in this list. This is asking for trouble, but I don't know of any alternatives, and it is still better than before where you would have to remember to update all those random static maps.

configdialog.cpp
```diff
@@ -82,7 +84,9 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 BOOL ConfigDialog::slider_changed(WPARAM wparam, HWND slider) {
 	int value = SendMessage(slider, TBM_GETPOS, 0, 0);
 
-	ConfigOption opt = reverse_lookup(slider_ids, GetDlgCtrlID(slider));
+	ConfigOption opt = *std::find_if(Opts::All.begin(), Opts::All.end(), [&](const ConfigOption &opt) {
+		return opt.dialog_control_id == GetDlgCtrlID(slider);
+	});
```
pierce-smith1, line 89: One example of things being slightly nicer because of this change. The code looks more complicated, but it allows us to completely get rid of this hacky `reverse_lookup` template and instead use a regular c++ algorithm directly.

spritecontrol.cpp
```diff
@@ -12,11 +12,11 @@ SpriteGenerator::SpriteGenerator() {
 	using namespace std::chrono;
 	std::srand(duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count());
 
-	PaletteGroup palette_group = (PaletteGroup) (cfg.at(YonkPalette));
+	PaletteGroup palette_group = (PaletteGroup) (cfg.at(Opts::Palette));
 
 	std::vector<const Palette *> bag_of_palettes;
 	if (palette_group == PaletteGroup::RandomlyGenerated) {
-		for (int i = (int) config_ranges.at(MaxColors).first; i < (int) config_ranges.at(MaxColors).second; i++) {
+		for (int i = Opts::MaxColors.range.first; i < Opts::MaxColors.range.second; i++) {
```
pierce-smith1: Another bit of code made much nicer.

configdialog.cpp
```diff
@@ -45,7 +45,9 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 			return TRUE;
 		}
 		case IDC_DEFAULTS_BUTTON: {
-			m_current_config = cfg_defaults;
+			for (const ConfigOption &opt : Opts::All) {
+				m_current_config[opt] = opt.default_;
```
pierce-smith1, line 49: This demonstrates what enumerating through all variants will look like if we do this strategy for other enums, like `PaletteName`. There is no raw for loop with nasty casting!

McAwesome123 (approved):
:lkapprove:

pierce-smith1:
Closing because the scope of this branch has gotten far away from the initial experiement here.

Issue #19, opened 2023-10-31
Make the dev experience of designing patterns not terrible


pierce-smith1:
There are a lot of pain points for making new patterns right now:

* The abstractions over patterns is perhaps a little too thin. Currently each pattern is just a function that takes in a sprite which you modify and then that's that. More complicated patterns, though, need more information, such as being able to access _other_ sprites, or the ability to defer applying changes to the sprites until all modifications have been calculated. We need a more flexible abstraction.
  * #20 attempts to change the current system so that the bridge between sprite-thinking behavior and the thinking functions is managed by a polymorphic `PatternPlayer` class. This is really complicated and I don't like it and I don't consider it a permanent solution.
* The coordinate system is insane. The coordinates of a sprite are NOT (0, 0) to (w, h), where w and h and width of the monitor(s), as you might expect. They are (-1.0, -1.0) to (1.0, 1.0), because that is how OpenGL expects them. This means, for instance, sprites appear _slower_ when they are moving up and down, since monitors are often wider than tall. We might want to abstract away this OpenGLism and present sprites as the coordinates we expect, so that making patterns is more natural.
* More?

PR #40, opened 2024-03-17
Replace enums for configs, palettes, and bitmaps

pierce-smith1:
Part of https://github.com/pierce-smith1/yok.scr/issues/30, but will not necessarily close it.

This is an extension of what we were doing in #39. The new "definition pattern" as an alternative to enums has been extended to palettes and bitmaps, and the pattern itself has been changed a little bit for config options to remove all hot-path accesses of `std::map`s. Code should be nicer and debug should run much, much faster.

We still need to do something about the `PatternName` enum, but that will probably need to be handled alongside whatever takes care of #19. 

This PR got mega enormous, so don't worry about being thorough. At a minimum you should just run it and make sure it not broke.


palettes.h
```diff
@@ -48,3 +56,768 @@ class RandomPalettes {
...
+	};
+
+	inline const static Definition Aemil = {
+		.name = L"aemil",
+		.group = PaletteGroup::Canon,
+		.data = new PaletteData {{
+			"#56eb8e",
+			"#84f5c3",
+			"#1d9550",
+			"#e29a56",
```
pierce-smith1, line 75: Cool, right?

bitmaps.h
```diff
@@ -0,0 +1,211 @@
...
+	Yokin,
+	Impostor,
+	YoyImpostor,
+};
+
+struct Bitmaps {
+	struct Definition {
+		auto operator<=>(const Definition &other) const = default;
+
+		std::wstring name;
```
pierce-smith1, line 30: Sadly I cannot make these fields `const`, because if I want this type to be copy-assignable, which is necessary to use in some `<algorithm>` stuff, they cannot be `const`.

bitmaps.h
```diff
@@ -0,0 +1,211 @@
...
+		.resource_id = IDB_BITMAP16,
+		.group = BitmapGroup::Impostor,
+		.data = load(IDB_BITMAP16),
+	};
+
+	inline const static Definition Vx = {
+		.name = L"vx",
+		.resource_id = IDB_BITMAP17,
+		.group = BitmapGroup::Impostor,
+		.data = load(IDB_BITMAP17),
```
pierce-smith1, line 154: Another unfortunate bit of necessary duplication between `.resource_id` and `.data`.

config.h
```diff
@@ -4,44 +4,172 @@
...
-	{ SpriteCount, L"SpriteCount" },
-	{ SpriteSize, L"SpriteSize" },
-	{ YonkShakeFactor, L"YonkShakeFactor" },
-	{ YonkPattern, L"YonkPattern" },
-	{ PatternChangeInterval, L"PatternChangeInterval" },
-	{ IsPatternFixed, L"IsPatternFixed" },
-	{ ImpostorChance, L"ImpostorChance" },
-	{ YonkPalette, L"YonkPalette" },
-	{ PlayOverDesktop, L"PlayOverDesktop" },
+struct Cfg {
```
pierce-smith1, line 13: I just cannot seem to decide on what the name of this type should be. I've decided on something very short since it's used all over the place, and I'm not quite happy with this but I'm OK with it.

palettes.h
```diff
@@ -19,15 +18,24 @@ enum PaletteIndex {
...
 public:
-	Palette(const std::array<Color, _PALETTE_SIZE> &colors);
-	Palette(const std::initializer_list<Color> &i_list);
+	PaletteData(const std::array<Color, _PALETTE_SIZE> &colors);
+	PaletteData(const std::initializer_list<Color> &i_list);
+	PaletteData(const std::array<std::string, _PALETTE_SIZE - 1> &hex_strings);
+};
+
+enum class PaletteGroup {
+	All,
```
pierce-smith1, line 29: Some enums like this one are really just tags, not stand-ins for things that should have been proper objects, and thus are spared from my crusade.

palettes.h
McAwesome123: :lkconcern:

common.h
```diff
@@ -23,4 +23,3 @@ enum Channel {
 	BLUE = 2,
 	ALPHA = 3
 };
-
```
McAwesome123, line 26: me when whitespace diff

spritecontrol.cpp
```diff
@@ -12,22 +12,22 @@ SpriteGenerator::SpriteGenerator() {
...
-		for (int i = (int) config_ranges.at(MaxColors).first; i < (int) config_ranges.at(MaxColors).second; i++) {
+		for (int i = Cfg::MaxColors.range.first; i < Cfg::MaxColors.range.second; i++) {
 			bag_of_palettes.push_back(RandomPalettes::random(i));
 		}
 	} else {
-		auto palette_names = PALETTES_BY_GROUP.at(palette_group);
-		for (auto name : palette_names) {
-			bag_of_palettes.push_back(&PALETTES.at(name));
+		auto group = PaletteGroups::get(palette_group);
+		for (const auto *palette : group.members) {
```
McAwesome123: This does not compile (C3535 and C2440).

McAwesome123 (changes requested):
Changes seem good but the project does not compile.

spritecontrol.cpp
pierce-smith1: Should be fixed now.

palettes.h
pierce-smith1: Well, uh... _I_ thought it was cool.

McAwesome123: y'know, seeing all the hex codes feels scary, but I don't think I have a problem with it.
it's probably better than having an array of arrays of four 0-255 values.

spritecontrol.h
```diff
@@ -11,13 +11,6 @@ const static float M_PI = std::acos(-1);
 
 using Sprites = std::vector<Sprite *>;
 
-struct Patterns {
-	struct Definition {
-		std::wstring name;
-		
-	};
-};
```
McAwesome123: I was wondering why this was like that.

palettes.h
```diff
@@ -48,3 +56,768 @@ class RandomPalettes {
...
+		.members = palettes_of_group(PaletteGroup::All),
+	};
+
+	inline static Definition get(PaletteGroup group) {
+		switch (group) {
+			case PaletteGroup::Canon: return Canon;
+			case PaletteGroup::NonCanon: return NonCanon;
+			case PaletteGroup::All: return All;
+		}
+	}
```
McAwesome123, line 812: It seems to not like the possibility of not returning a value (C4715), but I don't think it should matter too much?

McAwesome123 (changes requested):
I could be imagining it, but it feels like the amount of C4244 has gone down?

It mostly works fine, however, trying to use the All palette group crashes with a memory access violation.

palettes.h
pierce-smith1: My motivation was that it's easier to for example copy and paste color into and out of image editors. Plus it's way less ceremony to type. It is less safe because if you copy or type the hex code wrong it'll cause the program to blow up, but it should fail fast and be easy to spot the errors so I'm not _super_ worried about it.

palettes.h
pierce-smith1: Hmm yeah. That unsettles me a little bit but there's no meaningful result I can return if you pass in the random palette variant of that enum. I guess the reasonable thing to do is throw an exception, but we have 0 precedent for exceptions in this codebase, so idunno.

pierce-smith1: https://github.com/pierce-smith1/yok.scr/issues/41 I made an issue for us to think of some runtime error-checking strategy to use for this codebase.

palettes.h
pierce-smith1: #41 This is exactly the kind of thing some run-time sanity checks would guard against. Obviously if you type in a hex code wrong the program will just blow up, but it might not be obvious _what_ exactly you typed in wrong - this would help with that.

McAwesome123 (approved):
:lkapprove:

Issue #18, opened 2023-10-31
Solitaire style pattern


pierce-smith1:
Some kind of pattern where yokers fall and bounce on the bottom of the screen similar to how the cards bounce after winning a game of classic solitaire would be cute.

PR #42, opened 2024-03-19
Add Trails

McAwesome123:
yonkers trails!

currently, setting the trail length too high leads to slowdown. especially in debug, where it's comparable to when the config maps were causing lag. it's slightly less bad in release, but comparing max to none is still noticeably slower.

probably needs a lot (lotus) of cleanup. also need to actually add the config to the config window.

also, perhaps could be adapted in some way for #18?

### DIFF
--- a/Resource.rc
+++ b/Resource.rc
@@ -106,13 +106,14 @@ BEGIN
     LTEXT           "yokscr version ???",IDC_VERSION_LABEL,222,348,149,8,WS_DISABLED
     CONTROL         "Play over the desktop",IDC_PLAY_OVER_DESKTOP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,228,240,158,10
     LTEXT           "Trail Length:",IDC_STATIC,12,289,41,8
-    CONTROL         "",IDC_TRAIL_LENGTH,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,302,100,24
-    LTEXT           "Short",IDC_STATIC,12,308,18,8
-    LTEXT           "Long",IDC_STATIC,155,308,16,8
-    LTEXT           "Trail Spacing:",IDC_STATIC,12,332,44,8
-    LTEXT           "Near",IDC_STATIC,12,350,16,8
+    CONTROL         "",IDC_TRAIL_LENGTH,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,301,100,24
+    LTEXT           "Short",IDC_STATIC,12,307,18,8
+    LTEXT           "Long",IDC_STATIC,155,307,16,8
+    LTEXT           "Trail Spacing:",IDC_STATIC,12,331,44,8
+    LTEXT           "Near",IDC_STATIC,12,349,16,8
     CONTROL         "",IDC_TRAIL_SPACE,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,343,102,24
-    LTEXT           "Distant",IDC_STATIC,155,350,24,8
+    LTEXT           "Distant",IDC_STATIC,155,349,24,8
+    CONTROL         "Enable trails",IDC_TRAILS_ENABLED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,132,289,55,10
 END
 
 
--- a/Resource.rc
+++ b/Resource.rc
@@ -50,20 +50,20 @@ END
 // Dialog
 //
 
-DLG_SCREENSAVERCONFIGURE DIALOGEX 0, 0, 405, 316
+DLG_SCREENSAVERCONFIGURE DIALOGEX 0, 0, 405, 400
 STYLE DS_SETFONT | DS_MODALFRAME | DS_3DLOOK | DS_FIXEDSYS | DS_NOFAILCREATE | DS_CENTER | WS_POPUP | WS_CAPTION | WS_SYSMENU
 CAPTION "Yonkers"
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
-    DEFPUSHBUTTON   "OK",IDOK,348,295,50,14
-    PUSHBUTTON      "Cancel",IDCANCEL,294,295,50,14
+    DEFPUSHBUTTON   "OK",IDOK,348,379,50,14
+    PUSHBUTTON      "Cancel",IDCANCEL,294,379,50,14
     CONTROL         "",IDC_YONK_SHAKE_FACTOR,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,175,102,24
     LTEXT           "Shakiness:",IDC_STATIC,12,163,35,8
     LTEXT           "Population Size:",IDC_STATIC,204,18,52,8
     CONTROL         "",IDC_SPRITE_SIZE,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,133,102,24
     LTEXT           "Size:",IDC_STATIC,12,121,16,8
     LTEXT           "Impostor Amount:",IDC_STATIC,11,60,59,8
-    GROUPBOX        "Individuals",IDC_STATIC,6,109,186,180
+    GROUPBOX        "Individuals",IDC_STATIC,6,109,186,264
     GROUPBOX        "Family",IDC_STATIC,6,6,390,97
     LTEXT           "Starting Pattern:",IDC_STATIC,228,121,55,8
     GROUPBOX        "Animation ",IDC_STATIC,222,109,174,149
@@ -93,9 +93,9 @@ BEGIN
     LTEXT           "Pattern Change Frequency:",IDC_STATIC,228,199,90,8
     LTEXT           "Often",IDC_STATIC,228,217,20,8
     LTEXT           "Rarely",IDC_STATIC,360,217,22,8
-    PUSHBUTTON      "Defaults",IDC_DEFAULTS_BUTTON,222,295,50,14
+    PUSHBUTTON      "Defaults",IDC_DEFAULTS_BUTTON,222,379,50,14
     COMBOBOX        IDC_YONK_PATTERN,228,133,78,30,CBS_DROPDOWNLIST | CBS_SORT | WS_VSCROLL | WS_TABSTOP
-    LTEXT           "< https://doughbyte.com >",IDC_STATIC,222,277,90,8,WS_DISABLED
+    LTEXT           "< https://doughbyte.com >",IDC_STATIC,222,361,90,8,WS_DISABLED
     CONTROL         "Do not change",IDC_PATTERN_FIX,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,328,199,63,10
     LTEXT           "Population Diversity:",IDC_STATIC,11,18,67,8
     LTEXT           "None",IDC_STATIC,12,78,18,8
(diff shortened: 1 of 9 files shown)


sprite.cpp
```diff
@@ -68,8 +75,19 @@ void Sprite::transform() {
 	glScalef(m_size, m_size, 1.0f);
 }
 
+void Sprite::update_trail() {
+	if (round(cfg[Cfg::TrailLength]) <= 1) {
+		return;
+	}
+
+	trail.pop_back();
+	trail.insert(trail.begin(), Trail(m_texture, Point(final<X>(), final<Y>())));
```
pierce-smith1: ### unrelated to this PR's changes: i'm an idiot

While investigating performance concerns related to this vector manipulation, I made a very facepalm-inducing discovery. I didn't know VS could make flame graphs, but I stumbled into it by accident, and:

image

(this is Release)

If you don't know how to read a flame graph, each block is a function call, the blocks on top of a function call are the functions that function called, and wider blocks mean more time spent. And look, there are two almost identical structures in this graph, both of which have a base of `Yonker::update`. It turns out, every frame, we are calling `update` on every sprite **TWICE!** Once from `SpriteChoregrapher::update`,  and once from `Sprite::draw`.

The solution is to choose which one of these functions we should actually update the sprite in and remove the update from the other one. My vote is on removing it from `draw`, since I feel like the `draw` function should just be doing rendering, not logic.

Fixing this isn't related to yoker trails at all, but I figure we could just slip that into this PR, since it may just be one or two lines changed.

### ok, back on topic

Anyway, the other thing to take away from this flame graph is that the vector manipulation here does have a small but noticable effect on performance. You can't see it in the screenshot, but the blocks off to the right of `Yonker::update` are for `update_trail`. It's not a _huge_ amount of time, but it's notable. A big part of the reason for the slowness is probably the use of `insert`, which has to shuffle all the elements past the insertion position by one, which involves a lot of copies and is thus unnecessarily hard on memory. 

Now we can just decide not to care about this, which I wouldn't really argue with you much over. However:
* The effect of these vector operations is REALLY egregious in debug, and debug performance still matters
* There is (I think) a very elegant and satisfying alternative to `pop`ping and `insert`ing

What is that implementation? Well, there's a property about trails we can take advantage of here. That property is that trails are always the same length - they never gain or lose elements. This means, why should we even use `insert` or `pop` at all? Why not just change the elements in place?

So, here's my suggestion: use the `trail` vector as a _circular buffer_. Rather than shuffling the vector around so that the newest trail sprite is always at the beginning and the oldest at the end, keep a _start index_ that points to the oldest one. When you draw, you start iterating from that index, looping around once you reach the end of the vector, and finally stopping once you reach the index right before the start index (which you can call the "end index"). Then, in `update_trail`, instead of calling `pop_back` and `insert`, you just 1) replace the sprite at the start index with the newest one (this destroys the oldest sprite and replaces it with the newest sprite), and 2) move the start index forward by one! (this moves the start index to what is now the oldest sprite). Here's a little diagram of how it works: 

circ

So, do we actually want to do this? I think this is the right thing to do, but it might be kind of tricky. Personally, I think we should give it a shot, if for nothing else than the joy of failure.

sprite.cpp
```diff
@@ -68,8 +75,19 @@ void Sprite::transform() {
...
+}
+
 Yonker::Yonker(const Texture *texture, const Point &home) 
-	: Sprite(texture, home) { }
+	: Sprite(texture, home) {
+	m_emotion_vector = { 0.0f, 0.0f, 0.0f };
```
pierce-smith1: Nitpick: it's more efficient and idiomatic to initialize `m_emotion_vector` with a member initialization list rather than in the body of the constructor:
```c++
Yonker::Yonker(const Texture *texture, const Point &home) 
	: Sprite(texture, home), m_emotion_vector({ 0.0f, 0.0f, 0.0f }) { }
```

sprite.h
```diff
@@ -14,9 +14,13 @@
...
 #include "graphics.h"
 
+class Trail;
+
 class Sprite : public Identifiable<Empty> {
 public:
-	Sprite(const Texture *texture, const Point &home);
+	Sprite(const Texture *texture, const Point &home, const bool has_trail = true);
+
+	std::vector<Trail> trail;
```
pierce-smith1: This should be protected rather than public (I know why you made it public, but the next comment describes why it doesn't have to be), and nitpick: as a field it should be called `m_trail` instead of just `trail`.

scene.cpp
```diff
@@ -23,6 +23,13 @@ void Scene::draw() {
 	}
 
 	m_choreographer.update();
+
+	for (int i = (int)(round(std::clamp(cfg[Cfg::TrailLength], 1.0f, cfg[Cfg::MaxTrailCount] / cfg[Cfg::SpriteCount]) - 1) * max(round(cfg[Cfg::TrailSpace]), 1)) - 1; i >= 0; i -= (int) max(round(cfg[Cfg::TrailSpace]), 1)) {
```
pierce-smith1: Instead of rendering the trails here, we might want to do it inside the `Sprite`'s `draw` function. Then, instead of doing this huge for loop, we can just do:
```c++
for (const Trail &trail : m_trail) {
    m_trail.draw(ctx);
}
```
or, if we start using the trail vector as a circular buffer, then it might look more like:
```c++
for (size_t i = m_trail_start_index; i != m_trail_end_index; i = (i + 1) % m_trail.size()) {
    m_trail[i].draw(ctx);
}
```
This also means `m_trail` can be protected instead of public.

common.h
```diff
@@ -13,7 +13,7 @@ template <typename Base> class Identifiable : public Base {
 	}
 
 private:
-	const Id m_id;
+	Id m_id;
```
pierce-smith1, line 16: With the changes I described in the first comment, we may be able to put the const back on here!

sprite.cpp
```diff
@@ -8,8 +8,15 @@
 
 using std::get;
 
-Sprite::Sprite(const Texture *texture, const Point &home) 
-	: m_texture(texture), m_home(home), m_relpos(0.0f, 0.0f), m_size(cfg[Cfg::SpriteSize] / 1000.0f) { }
+Sprite::Sprite(const Texture *texture, const Point &home, const bool has_trail)
+	: m_texture(texture), m_home(home), m_relpos(0.0f, 0.0f), m_size(cfg[Cfg::SpriteSize] / 1000.0f)
+{
+	if (has_trail && round(cfg[Cfg::TrailLength]) > 1) {
+		for (int i = 0; i < (int)(round(std::clamp(cfg[Cfg::TrailLength], 1.0f, cfg[Cfg::MaxTrailCount] / cfg[Cfg::SpriteCount]) - 1) * max(round(cfg[Cfg::TrailSpace]), 1)); i++) {
```
pierce-smith1: I know I'm not exactly a good example for not writing way-too-long single lines of code, but this is a pretty beefy for loop - it wraps three whole times in the preview here. It would help a lot to break out into some variables so it's clearer what each of these calculations is doing and saves some space.

sprite.h
```diff
@@ -62,6 +67,15 @@ class Impostor : public Sprite {
 public:
 	Impostor(const PaletteData *palette, const Point &home);
 
+	virtual void update(Context &ctx) override;
+
 protected:
 	static Bitmaps::Definition &random_bitmap();
 };
+
+class Trail : public Sprite {
```
pierce-smith1: The name of this class is a little misleading, as it doesn't represent the whole trail, just a single sprite in the trail. Maybe it should be called `TrailSprite` or something.

pierce-smith1 (commented):
You are about to read a lot of words, but overall I'm really happy about this, the high-level implementation is really well thought out and executed. This is not an easy problem and I'm surprised you got it working to begin with.

sprite.cpp
McAwesome123: It sounds like a pretty good idea and I have actually thought of it, just the difficulty of implementing it compared to the current solution has somewhat dissuaded me.

sprite.h
McAwesome123: I have a feeling that private, protected, and public are forever gonna be one of those things that you **have** to tell me to use properly otherwise I just won't.

scene.cpp
McAwesome123: that's probably reasonable, the main reason I did it this way is because of sprite layering

pierce-smith1: Shit I hadn't considered that. I think we could still do something cleaner by introducing a `draw_trail` function for the `Sprite` though, that would be another way to get rid of this for loop and also keep the trail protected instead of public.

sprite.cpp
McAwesome123: Unsurprisingly, I agree. There's far too much happening in this one line.

sprite.cpp
McAwesome123: I have made an attempt at implementing something along those lines

sprite.cpp
McAwesome123: This has been done.

common.h
McAwesome123: I forgot about it before making the commit, but it can't be added back yet anyway because my method of replacing elements still doesn't seem compatible with it.

sprite.cpp
McAwesome123: These have been cleaned up some.

sprite.h
McAwesome123: It's been renamed.

sprite.h
```diff
@@ -14,9 +14,14 @@
...
 
+class TrailSprite;
+
 class Sprite : public Identifiable<Empty> {
 public:
-	Sprite(const Texture *texture, const Point &home);
+	Sprite(const Texture *texture, const Point &home, const bool has_trail = true);
+
+	std::vector<TrailSprite> m_trail;
+	int m_trail_iterator;
```
pierce-smith1: Agh, please don't call this an "iterator", you almost scared me into thinking you were actually storing an iterator. This is a start index, so call it that.

Also as an index, it should be a `size_t`.

Also also, like `m_trail`, it should definitely be protected.

sprite.h
```diff
@@ -62,6 +68,18 @@ class Impostor : public Sprite {
...
 	static Bitmaps::Definition &random_bitmap();
 };
+
+class TrailSprite : public Sprite {
+public:
+	TrailSprite(const Texture *palette, const Point &home);
+
+	virtual void update(Context &ctx) override;
+
+	static int get_trail_length();
```
pierce-smith1, line 85: Cool.

config.h
```diff
@@ -142,6 +141,36 @@ struct Cfg {
...
+	inline const static Definition TrailSpace = {
+		.index = __COUNTER__,
+		.name = L"TrailSpace",
+		.default_ = 10.0f,
+		.range = { 1.0f, 50.0f },
+		.dialog_control_id = IDC_TRAIL_SPACE,
+	};
+
+	// aka: kindly, calm the fuck down mate
+	inline const static Definition MaxTrailCount = {
```
pierce-smith1, line 161: If we aren't going to let users change this (which I don't think we should), it shouldn't be a config option, it should just be a const static variable somewhere.

Resource.rc
```diff
@@ -93,18 +93,27 @@ BEGIN
...
     CONTROL         "Play over the desktop",IDC_PLAY_OVER_DESKTOP,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,228,240,158,10
+    LTEXT           "Trail Length:",IDC_STATIC,12,289,41,8
+    CONTROL         "",IDC_TRAIL_LENGTH,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,301,100,24
+    LTEXT           "Short",IDC_STATIC,12,307,18,8
+    LTEXT           "Long",IDC_STATIC,155,307,16,8
+    LTEXT           "Trail Spacing:",IDC_STATIC,12,331,44,8
+    LTEXT           "Near",IDC_STATIC,12,349,16,8
+    CONTROL         "",IDC_TRAIL_SPACE,"msctls_trackbar32",TBS_BOTH | WS_TABSTOP,48,343,102,24
+    LTEXT           "Distant",IDC_STATIC,155,349,24,8
+    CONTROL         "Enable trails",IDC_TRAILS_ENABLED,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,132,289,55,10
```
pierce-smith1, line 116: image

UX: If trails are disabled, the sliders should be disabled too. Letting people twiddle sliders that do nothing is really confusing. Look at the pattern fixed checkbox and the pattern duration slider to tell how I did this in the past.

pierce-smith1 (approved):
I think everything else about this is good. I played with it a little bit and couldn't find anything else super concerning, I trust your validation better than mine anyway, so just fix those other two things and you can merge this in whenever you want. I like it, you did an awesome job.

config.h
McAwesome123: The idea here was that it doesn't appear in the config but you can still edit it in the registry _if_ it needs to be changed for whatever reason.

Issue #10, opened 2023-10-24
Take a look at all the possible loss of data warnings


pierce-smith1:
All the C4244 and C4267s are making compilation very noisy. I have no idea if any of them speak to an actual issue - they probably don't, since everything has been going smoothly as far as I can tell - but we should do our due diligence, and get rid of all of them while we're at it.

McAwesome123:
Can't say much about how good of a solution it is, but using explicit casts seems to get rid of the warning (#23)

pierce-smith1:
That's far from solving all of them though.

McAwesome123:
This feels more like "muting" them than solving them, just without _literally_ muting them

Issue #41, opened 2024-03-17
Introduce a standard for run-time error checks


pierce-smith1:
My philosophy so far on this project has been "don't write shit down wrong," which has served us pretty well so far, but because of the wild goose chase that C++ errors are, maybe it's time to at least think about doing some actual run-time sanity checking to make catching these mistakes easier. 

We'll need to come up with a standard for doing this. Some options:

* Throw an exception - this is nice because it should trigger the debugger and give us a useful stack trace.
* Use some kind of debug-only assert statement - this is nice because we don't pay the costs of it in release, though the "costs" of any other strategy will probably be negligible. It's also nice because it won't cause the screensaver to just randomly quit in release if there is a bug, but if we have a bug the screensaver is probably just going to crash anyway.
* Log something to the console - probably the worst choice because it doesn't really do anything useful that a debugger doesn't already.

PR #43, opened 2024-03-21
Destroy all warnings

pierce-smith1:
#10 

Shut up shut up shut up shut up.

As part of this work, since it made things easier, I also converted every float in our codebase to a double. Originally I thought we couldn't do this because OpenGL required floats, but I was simply incorrect, it can work with doubles just fine.

I have also set warnings to be treated as errors in Release mode, to help gently encourage us to not introduce any more.

We're on /W3 right now. There are higher warning levels, but we will probably stay at /W3. I have absolutely no intention of increasing us to the most pedantic warning settings because we would have hundreds and hundreds of them, and these loss of data warnings we're dealing with now are already pretty fucking pointless. It's a shame because I would love to get good warnings that can catch actual problems, but they will be buried by 1000 other warnings that might be "oh shit" in other codebases but are "tell someone who cares" in this one. Maybe we need to curate a warning list or something.

### DIFF
--- a/common.h
+++ b/common.h
@@ -23,3 +23,7 @@ enum Channel {
 	BLUE = 2,
 	ALPHA = 3
 };
+
+template <typename To, typename From> To cast(From value) {
+	return static_cast<To>(value);
+}
--- a/palettes.h
+++ b/palettes.h
@@ -4,6 +4,8 @@
 #include <set>
 #include <initializer_list>
 #include <vector>
+#include <string>
+#include <stdexcept>
 
 #include "common.h"
 
@@ -51,8 +53,8 @@ private:
 	static Color random_gray();
 	static Color darken_color(const Color &color);
 	static Color lighten_color(const Color &color);
-	static Color noisify(const Color &color, float degree = 1.0f);
-	static Color recolorize(const Color &color, float red_weight, float green_weight, float blue_weight);
+	static Color noisify(const Color &color, double degree = 1.0);
+	static Color recolorize(const Color &color, double red_weight, double green_weight, double blue_weight);
 	static std::set<GenerationTraits> random_traits();
 };
 
@@ -808,6 +810,7 @@ public:
 			case PaletteGroup::Canon: return Canon;
 			case PaletteGroup::NonCanon: return NonCanon;
 			case PaletteGroup::All: return All;
+			default: throw std::domain_error("PaletteGroup " + std::to_string((int) group) + " is not valid.");
 		}
 	}
 };
--- a/sprite.cpp
+++ b/sprite.cpp
@@ -19,7 +19,7 @@ void Sprite::draw(Context &ctx) {
 	update(ctx);
 	m_texture->apply();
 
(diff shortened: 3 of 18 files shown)


common.h
```diff
@@ -23,3 +23,7 @@ enum Channel {
 	BLUE = 2,
 	ALPHA = 3
 };
+
+template <typename To, typename From> To cast(From value) {
+	return static_cast<To>(value);
```
pierce-smith1, line 28: Pure convenience for writing a static_cast with less typing because we have to do it somewhat frequently. This probably wasn't necessary and might actually be slightly harmful in the long term for reasons I am too tired to explain but whatever.

I actually didn't have to static cast as much as I thought I would need to, which is good.

yokscr.vcxproj
```diff
@@ -85,6 +85,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <TargetExt>.scr</TargetExt>
     <TargetName>yok</TargetName>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
```
pierce-smith1, line 88: This doesn't do anything, false is the default here, this is just what I get for playing with the project settings too much.

palettes.h
```diff
@@ -808,6 +810,7 @@ struct PaletteGroups {
 			case PaletteGroup::Canon: return Canon;
 			case PaletteGroup::NonCanon: return NonCanon;
 			case PaletteGroup::All: return All;
+			default: throw std::domain_error("PaletteGroup " + std::to_string((int) group) + " is not valid.");
```
pierce-smith1, line 813: I am not necessarily making a decision on #41 here, I just wanted to do something semi-reasonable to suppress the missing branch error.

sprite.cpp
```diff
@@ -52,9 +52,9 @@ void Sprite::update(Context &ctx) {
 		}
 	};
 
-	float edge_boundary = 0.15f + m_size / 1.1f;
-	float horizontal_correction = max((float) ctx.rect().right / (float) ctx.rect().bottom, 1.0f);
-	float vertical_correction = max((float) ctx.rect().bottom / (float) ctx.rect().right, 1.0f);
+	double edge_boundary = 0.15f + m_size / 1.1f;
+	double horizontal_correction = max((double) ctx.rect().right / (double) ctx.rect().bottom, 1.0f);
+	double vertical_correction = max((double) ctx.rect().bottom / (double) ctx.rect().right, 1.0f);
```
McAwesome123: I don't know if this is intentional or a mistake, but you are still using floats here

McAwesome123 (commented):
I have yet to try building/running this but for now it seems good.

McAwesome123 (approved):
Seeing a build output with no warnings feels unnatural.

pierce-smith1:
This is waiting for #42.

pierce-smith1:
Now it's just waiting for whenever, I need to go through the code that has gone into master since this was made but I'm lazy.

Issue #28, opened 2024-01-06
Advanced palette selection screen


pierce-smith1:
What if you could fine-tune the palettes you wanted to a frankly irresponsible degree of detail? Letting you see all the palettes in a group, choose the individual weights, turn ones on and off, mix and match from different groups, etc. This would be very useful to #26 and #27.

Some UI sketching needs to be done first.

pierce-smith1:
This has been partially addressed by #44. We might decide that's good enough...

Issue #27, opened 2024-01-06
Custom palette generation


pierce-smith1:
A while ago I made [this thing](https://doughbyte.com/aut/_palgen/), why not basically port it to yok.scr and let people design their own palettes?

PR #44, opened 2024-03-24
Custom palettes

pierce-smith1:
> [!Caution]
> @McAwesome123 Because of the way I decided to resolve merge conflicts, this might explode or do weird shit when you try to run it locally. I accidentally repurposed the index of one of the trail config options for a new one I added in this PR. I would recommend you just delete your entire yokscr registry key before trying to run this.

#27 and part of #28.

image

You can now create your own palettes! 

This PR adds a config option, "Use custom palette set," that replaces the normal palettes with ones you define. It also adds a GUI tool for designing your own palettes which will be used if you choose that option.

All custom palettes are stored directly in the registry. As you create palettes, you will make `_CustomPalette:` entries that store a semicolon-separated list of colors. A comma-separated list of all custom palette names is stored in `_CustomPalette_index`.

This is a pretty big PR, but unlike my last ones it's mostly new code and not just refactoring, so it might actually be worth it to read.

Some concerns:
* I have tried to protect against putting nasty characters and other bad stuff in the registry via the palette names, but I may not have fully locked that down. More rigorous testing might be needed.
* Some of the UI controls do some kind of janky shit and I'm surprised it works at all on my machine. I can't wait to find out it doesn't work on other computers.
* Are there performance issues?????


spritecontrol.cpp
```diff
@@ -14,24 +14,31 @@ SpriteGenerator::SpriteGenerator() {
...
+		bag_of_palettes = group.members;
+	}
+
+	if (bag_of_palettes.empty()) {
+		bag_of_palettes = PaletteGroups::All.members;
 	}
 
+	int min_colors = bag_of_palettes.size() < Cfg::MaxColors.range.first
+		? bag_of_palettes.size()
+		: Cfg::MaxColors.range.first;
```
pierce-smith1: Normally you can't have less than 2 palettes, but I wanted to let people run with just a single palette if they really wanted to, AND to have a reasonable default if someone just defined no palettes at all. These extra checks to `bag_of_palettes.size()` protect against that.

palettes.h
```diff
@@ -811,3 +814,51 @@ struct PaletteGroups {
...
+	std::vector<Palettes::Definition> get_all_custom_palettes();
+
+private:
+	std::wstring serialize(const PaletteData &palette);
+	PaletteData deserialize(const std::wstring &serialized);
+
+	RegistryBackedMap m_map;
+};
+
+class PaletteGroupRepository {
```
pierce-smith1, line 839: Originally, I was going to put support for multiple custom palette groups in this PR, but that made the UI almost incomprehensible, so I abandoned it for now. I'm keeping this class around because I might come back to it. It's fully functional (at least I think so, as confidently as I can be without automated testing..!)

config.cpp
```diff
@@ -41,33 +44,44 @@ Registry::Registry()
...
 }
 
 float Registry::get(const std::wstring &opt, float default_) {
-	wchar_t result[1 << 6] {};
+	auto value = get_string(opt, std::to_wstring(default_));
+	return std::stof(value);
+}
+
+std::wstring Registry::get_string(const std::wstring &opt, const std::wstring &default_) {
+	wchar_t result[1 << 12] {};
```
pierce-smith1, line 82: I'm too tired to fix this right now but this looks like a free stack smash to me... this needs to get looked at again.

config.cpp
```diff
@@ -41,33 +44,44 @@ Registry::Registry()
 	);
 }
 
+Registry::~Registry() {
+	RegCloseKey(m_reg_key);
+}
```
pierce-smith1, line 49: The `Registry` class was pretty weird. It never closed the key it created, but it also wasn't a singleton, so it was extremely unclear where and when it was safe to use. I made it close the key once destructed so it has more RAII/transactional-ish semantics.

pierce-smith1:
image

I have really bad habits, never name your commits like this.

McAwesome123:
> image
> 
> I have really bad habits, never name your commits like this.

lkfear

McAwesome123 (approved):
Very cool.

pierce-smith1:
I'm afraid of this but it's going out.

PR #55, opened 2024-03-29
Add predefined palettes in the palette customizer

### LINKED ISSUE #51: Allow users to add existing palettes to the custom palette set
Custom palettes are cool, but making palette sets would be more expressive and interesting if you could also throw some pre-defined ones in the mix. Not only could you then effectively make custom remix palettes (think something like 10 aemils and 1 loxxe), you could also build your custom palettes off of existing ones. It's probably one of the most effort-efficient bonuses to custom palettes we can introduce!

This would get us one step closer to a full scale palette customizer like described in #28, and honestly it might be enough to close that issue too. Would we really need much more than this?

### LINKED ISSUE #52: Palette customizer allows you to create another palette of the same name
When you add a new palette in the palette customizer, you are prompted to give it a name. You are allowed to give it the same name as a palette that already exists, in which case the existing one will be destroyed without warning. You shouldn't be allowed to do this.

pierce-smith1:
Closes #51 and #52.

image

This PR introduces an "Add built-in palette" button to the palette customizer which will let you, well, do that. I also fixed #52 while I was at it - now if you try to add a palette with a name that already exists, it will be called "palette 2", or "palette 3", and so on.

On the dev side, I also introduced an abstraction for easily getting a c++ string from an operation that normally requires a lengthy buffer allocation ceremony, since this shit is EVERYWHERE in Win32. This should make interacting with windows a lot nicer and safer.


McAwesome123 (approved):
Seems good

PR #56, opened 2024-03-29
Add palette sharing in the palette customizer

### LINKED ISSUE #48: Add a way to share palettes
Anything that doesn't involve clicking yes on a message box like this.
image

Two possible ways of achieving this include generating a string describing the palettes that is then either exported to a file or simply copied to the clipboard.

### LINKED ISSUE #54: The Red/Green/Blue input boxes on the custom color dialog don't work right
According to @McAwesome123 (because I haven't tried to replicate this yet), the input boxes that let you specify RGB as 0-255 directly only let you enter 0-99. If you have more details to add that would be cool.

I think this bug is fairly low priority, but it should be documented.

McAwesome123:
image

This would let you easily share your palettes with others without them having to edit the registry.
Window design may or may not be subject to change. (For example: the export button may be removed as it isn't strictly necessary. Especially if I can't figure out how to make it functional.)

This should resolve #48 and fix #54.


configdialog.cpp
```diff
@@ -294,6 +294,20 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 			refresh_palette_list();
 			break;
 		}
+		case IDC_PALDLG_IMPORT_EXPORT_PALETTES: {
+			std::wstring *palettes = (std::wstring *) DialogBoxParam(
+				NULL,
+				MAKEINTRESOURCE(DLG_IMPORT_EXPORT_PALETTES),
+				m_dialog,
+				(DLGPROC) ScreenSaverImportExportPalettesDialog,
+				(LPARAM) (export_palettes().c_str())
```
pierce-smith1: ⚠️ Because the result of `export_palettes()` is not assigned to anything, it is a temporary object and is destroyed immediately. `c_str()` returns a pointer to the string's underlying storage, but by the time this expression finishes evaluating, the string will have been destroyed and the pointer will be invalid. 
The result of `export_palettes` should be stored in a variable here so it has somewhere to live and won't instantly be killed.

The safest (but not necessarily most efficient) thing to do here is probably to use the heap. This also makes it symmetric with how `EndDialog` works, which is nice.
```c++
auto *exported_palette = new std::wstring(export_palette());
std::wstring *palettes = (std::wstring *) DialogBoxParam(
    NULL,
    MAKEINTRESOURCE(DLG_IMPORT_EXPORT_PALETTES),
    m_dialog,
    (DLGPROC) ScreenSaverImportExportPalettesDialog,
    (LPARAM) exported_palette
);

... later, in the dialog proc

case WM_INITDIALOG: {
    auto *exported_palette = (std::wstring *) lparam;
    Edit_SetText(
        palettes_input,
	exported_palette->c_str()
    );
    delete exported_palette;
    return TRUE;
}
```

configdialog.cpp
```diff
@@ -518,6 +532,23 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+	auto all_palettes = m_palette_repo.get_all_custom_palettes();
+	std::wstring palettes_string = L"";
+
+	for (const auto &palette : all_palettes) {
+		auto name = palette.name;
+		auto *data = palette.data;
+
+		palettes_string.append(palette.name + L": ");
+		palettes_string.append(m_palette_repo.serialize(*data));
+		palettes_string.append(L"\r\n");
```
pierce-smith1: Line breaks are a text encoding nightmare and it makes me very anxious to put them in something that will be pasted around and sent over the internet. The clipboard could fuck it up, someone could paste it into an editor that replaces \r\n with \n, someone could paste it into discord and that replaces \r\n with \n, it's just bad. I suggest we pick a different separator.

Now this of course means your export code is all one line, which is a lot uglier, but that might actually be a good thing. It's nice for a data transfer code to be indecipherable nonsense so that people treat it like an opaque box of horrors they don't want to touch. If people look at and they think they know how it works, they might be empowered to touch it or reformat it so it looks nicer, and then whoops they fucked it up.

(To that end you could also get rid of the space after the palette name, since I think all it's doing is just making it look nicer)

configdialog.cpp
```diff
@@ -518,6 +532,23 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+	}
+	palettes_string.append(L"~");
```
pierce-smith1: Why the squiggle at the end here?

configdialog.cpp
```diff
@@ -703,3 +734,57 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
...
+							// The only catastrophe I can directly forsee is commas, but may as well just only allow
+							// alphanumerics and a few special characters to be extra safe.
+							static std::wstring valid_chars = L"qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM0123456789 -_!?:#;~\r\n";
+							std::wstring palettes = palettes_buffer;
+
+							bool palettes_too_short = palettes.size() < PaletteCustomizeDialog::MinPaletteNameSize + 57;
+							bool palettes_has_invalid_chars = palettes.find_first_not_of(valid_chars) != std::wstring::npos;
+
+							HWND ok_button = GetDlgItem(dialog, IDOK);
+							EnableWindow(ok_button, !palettes_too_short && !palettes_has_invalid_chars);
```
pierce-smith1: I can't tell if this is incomplete or not so I'll just mention it, but this validation pass is going to need a lot more to ensure that the code someone pasted in is actually importable. You will have to check that the code is in _exactly_ the same format as the string produced by `export_palettes`. So, it has a palette name followed by a colon followed by seven hex codes followed by another separator... and so on.

Because that is likely going to be a complex validation, it's definitely worth splitting off into its own function somewhere.

pierce-smith1:
We should also think about how the import is actually going to work. Is it going to replace all of their existing palettes with the imported ones, or is it going to add each of the imported palettes as new ones? If it does something destructive, the dialog should absolutely include a warning that it's going to delete all your palettes or whatever.

configdialog.cpp
McAwesome123: yeah, fair enough
given this, it's worked surprisingly well so far.

configdialog.cpp
McAwesome123: They pretty much are there _just_ to make it look nicer. I don't intend on making them required.

That being said, windows really does not like this having just \n

configdialog.cpp
McAwesome123: Felt like it. (Might use it as a check to see if the end has been reached? Idk there's probably better ways of doing that.)

configdialog.cpp
McAwesome123: Yes and no.
It's complete in the sense that I don't plan on adding more validation _here_ but incomplete in the sense that there _will_ be more validation. It will just be a bit more verbose than graying out a button.

McAwesome123:
> We should also think about how the import is actually going to work. Is it going to replace all of their existing palettes with the imported ones, or is it going to add each of the imported palettes as new ones? If it does something destructive, the dialog should absolutely include a warning that it's going to delete all your palettes or whatever.

I'd prefer to go for adding each of the imported palettes as new ones rather than replacing the existing ones.

configdialog.cpp
```diff
@@ -518,6 +535,78 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		palettes_string.append(palette.name + L": ");
+		palettes_string.append(m_palette_repo.serialize(*data));
+		palettes_string.append(L";\r\n");
+	}
+	palettes_string.append(L"~");
+
+	return palettes_string;
+}
+
+void PaletteCustomizeDialog::import_palettes(std::wstring *palettes) {
```
pierce-smith1: Nitpick: this should take in a string via const reference. There's no reason for it to be a pointer, and we definitely don't want to accidentally modify it so it should be const.

configdialog.cpp
```diff
@@ -518,6 +535,78 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+	static std::wstring valid_chars = L"qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM0123456789-_!?:";
+	static std::wstring valid_hex_chars = L"0123456789abcdefABCDEF";
```
pierce-smith1: Nitpick: both of these statics should be const. We don't want them to change.

configdialog.cpp
```diff
@@ -518,6 +535,78 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+				}
+				current_data[i] = palettes->substr(current_index, 6);
+				if (current_data[i].find_first_not_of(valid_hex_chars) != std::wstring::npos) {
+					throw ERROR_BAD_ARGUMENTS;
+				}
+				current_index = palettes->find_first_of(L';', current_index);
+			}
+			last_successful_palette_name = current_name;
+			if (palettes->find_first_of(valid_chars, current_index) == std::wstring::npos) {
+				if (palettes->find_first_of(L'~', current_index) != std::wstring::npos) {
```
pierce-smith1: I'm like, 95% certain this terminating character doesn't do anything useful. The end of the palette data _is_ the end of the string - trying to check that a string always ends with some character does not add any additional information. If you hit the end of the string and your parser is expecting another palette, you know you're done. If it's expecting anything else, like another byte of a color code, then the data is bad.

configdialog.cpp
```diff
@@ -518,6 +535,78 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+				break;
+			}
+			else {
+				current_index = palettes->find_first_of(valid_chars, current_index);
+			}
+		}
+		if (!graceful_exit) {
+			MessageBox(
+				NULL,
+				(L"A terminating character was not found. One or more palettes may have failed to import. Last successful import: " + last_successful_palette_name).c_str(),
```
pierce-smith1: It looks like the intent of this code is to convert and save as many palettes as possible before running into an error. I argue that this should be an all-or-nothing operation - either **every** palette converts and is saved, or **none** of them are.

My rationale: If a user sees a message saying "whoops, we converted some of your palettes, but not all of them!" they now have to _turn on their brain_ to figure out what to do with that information. Should I delete the ones that converted successfully before and try again? Which ones even converted successfully? Now I have to look at the message box and see which ones came before the last successful import! But shit, I already closed the message box... and crap, I have all these palettes in my list, I can't even remember which ones I was trying to import! How was I supposed to know that anyway, all I did was copy some random thing someone sent me! Gosh, I feel stupid, computers just aren't for me :(

Meanwhile, if the message box just said "It didn't work," then their only options are to try copy and pasting again or complain to someone. These were effectively their only options anyway, but now it's obvious. The world is infinitely simpler.

configdialog.cpp
```diff
@@ -518,6 +535,78 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+				L"Palette Customizer",
+				MB_ICONWARNING
+			);
+		}
+	}
+	catch (...) {
+		MessageBox(
+			NULL,
+			(L"An error occurred when importing palette \"" + current_name + L"\". Last successful import: \"" + last_successful_palette_name + L"\"").c_str(),
```
pierce-smith1: To expand on the last comment, I think it should be possible to avoid showing users an error message altogether. Error boxes are annoying - they make a stupid noise, rip control away from your program, say some dumb technical jargon that means nothing to you, and then leave you confused what to do with the information it tried to tell you but you probably didn't read because it's been a long day and you're tired and you just want the fucking god forsaken thing to do the fucking thing it's fucking supposed to.

In my opinion, if the import is going to fail, then you just shouldn't be able to push the button at all. The Import button should only be active if the string you pasted in is OK. So, this function could be adjusted to just return a "bad" signal every time it would throw, and if everything goes alright then it returns the palettes that it parsed. Then rather than being called whenever someone pushes "Import," it's called every time the input box changes. If the string is bad, disable the import button - but if it parsed successfully, then store the results somewhere so we can save them when the user finally does click Import. 

(This is _exactly_ the use case for `std::optional`. I think a perfect return type for this function might be `std::optional<std::vector<PaletteData>>`.)

pierce-smith1 (commented):
As far as the parsing/validating code itself goes, it looks pretty solid. I wasn't able to break it and it looks safe. But I do have some thoughts on _how_ we're using it.

configdialog.cpp
```diff
@@ -544,66 +544,259 @@ std::wstring PaletteCustomizeDialog::export_palettes() {
...
 }
 
 void PaletteCustomizeDialog::import_palettes(std::wstring *palettes) {
-	static std::wstring valid_chars = L"qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM0123456789-_!?:";
-	static std::wstring valid_hex_chars = L"0123456789abcdefABCDEF";
-	size_t current_index = palettes->find_first_of(valid_chars.c_str());
-	std::wstring last_successful_palette_name;
-	std::wstring current_name;
-	std::wstring current_data[7];
+	using namespace std;
```
McAwesome123: evil.

configdialog.cpp
```diff
@@ -544,66 +544,259 @@ std::wstring PaletteCustomizeDialog::export_palettes() {
...
+
+	const enum class Errors {
+		none = 0,
+		name_too_short = 1 << 0,
+		name_too_long = 1 << 1,
+		name_invalid_characters = 1 << 2,
+		colors_too_few = 1 << 3,
+		colors_incorrect_length = 1 << 4,
+		colors_invalid_characters = 1 << 5,
+	};
```
McAwesome123: This feels like an appropriate use of enums, no?

configdialog.cpp
```diff
@@ -805,17 +998,17 @@ LRESULT CALLBACK ScreenSaverImportExportPalettesDialog(HWND dialog, UINT message
 			delete exported_palettes;
 			return TRUE;
 		} case WM_COMMAND: {
-			wchar_t palettes_buffer[1 << 16] { L'\0' };
+			std::unique_ptr<wchar_t> palettes_buffer(new wchar_t[1 << 16](L'\0'));
```
McAwesome123: I didn't like the warning of having 128KB+ of data on the stack, so I tried turning this into a unique pointer.

configdialog.cpp
```diff
@@ -518,6 +535,271 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		}
+		palettes_string.append(colors);
+		palettes_string.append(L"\r\n");
+	}
+	palettes_string.append(L"~");
+
+	return palettes_string;
+}
+
+void PaletteCustomizeDialog::import_palettes(std::wstring *palettes) {
```
pierce-smith1, line 560: This function is over 200 lines long. Frankly, that is ludicrous. I understand that a lot of complexity comes from doing the in-depth error reporting, but I am worried this has been over-engineered. 

Now I can tell a lot of love and care went into this, so it probably works fine, and the last thing I want you to do after solving this hard problem is tell you to go throw it all out. But this is one of the reasons I like to not solve hard problems. This problem would have been much easier without detailed error reporting - it might have been just a few calls to `split` and few checks for vector size to do basic sanity checks.

Again, at this point all I care about is that it works, so my "real" review of this once it is closer to done will not really be a line-by-line thing but a suite of runtime tests. Please don't waste a bunch of time simplifying it now, just keep going with actual importing and we'll come back to this if we find out we need to.

pierce-smith1: This function is over 200 lines long, with lines that have up to seven levels of indentation, a dozen random variables with confusing names declared up front (why is something called `is_valid` a vector of integers? what is valid? what values of that type represent validity?) - frankly it is a mess. I understand that a lot of complexity comes from doing the in-depth error reporting, but I am worried that even then this has been wildly over-engineered. It is a very easy mistake to make, because programming and solving problems and being clever is fun, but it is a mistake nonetheless. We are left with an enormous, nearly untouchable bird's nest that I don't even want to try to understand. Am I going to be able to come into it and fix bugs if I need to? Are you going to be able to come back to this and understand it enough to change it when our palette code format changes 6 months later?

Now I can tell a lot of love and care went into this, so it probably works fine, and the last thing I want you to do after solving this hard problem is tell you to go throw it all out. So let's not throw it out right now. But this is one of the reasons I like to not solve hard problems if I don't have to - general implementation decisions aside, it would have been much easier and more graceful and maintainable without the detailed error reporting, it might have been just a few calls to `split` and few vector size checks to do basic sanity checks. And seriously, what errors are we _really_ expecting from copying and pasting some text?

Again, at this point all I care about is that it works, so my "real" review of this once it is closer to done will not really be a line-by-line thing but a suite of runtime tests. Please don't waste a bunch of time simplifying it now, just keep going with actual importing and we'll come back to this if we find out we need to.

(It's worth acknowledging that I am NOT clean of the 300-lines-of-unreadable-bullhsit sin, I'm sure you remember #20 and the horrors I created there. And like with that one, it's not always practical to fix the bullshit, but we should still feel ashamed.)

configdialog.cpp
pierce-smith1: There are some things you should turn to before going nuclear with `using namespace std`:

* Making type aliases for your longest and most annoying types, e.g. `using RawColors = std::vector<std::array<std::wstring, 7>>`
* Use only exactly what you need: `using std::wstring; using std::vector; using std::array;`
* Declaring with `auto` wherever possible

The main reason I don't like `using namespace std` and friends, except for truly deranged namespaces like `std::chrono` or `std::string_literals`, is because `using namespace` completely destroys all of the safety provided by namespaces. The `std` namespace contains A LOT of crap, much of it with really simple names, like `begin` or `get`. If I write a function name or class name or even a variable name(!) that accidentally collides with something in that namespace, what the hell happens? I don't know and I don't want to find out. 

That being said, you did do the right thing by containing the `using` to just this function.

configdialog.cpp
```diff
@@ -518,6 +535,271 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+	using namespace std;
+
+	const static wstring valid_chars = L"qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM0123456789-_!?:";
+	const static wstring valid_hex_chars = L"0123456789abcdefABCDEF";
+	const static size_t colors_amount = max(7, (size_t)ceil(PaletteCustomizeDialog::MaxPaletteNameSize / 8.0));
```
pierce-smith1: Why is this ever not 7?

configdialog.cpp
pierce-smith1: Is this appropriate for an enum? It's debatable. It violates the "spirit" of an enum, which is that the variants are mutually-exclusive, but C++ enums are more or less just fancy names for integers and not much else is enforced, so eh, whatever.

However, you are `static_cast<uint_8>`ing like a motherfucker in order to treat these enum values like bit flags, and that is messying up the code quite a lot. I _think_ you can do something like this to solve that though:

```c++
enum ColorParseError : uint8_t {
    stuff = 1 << 0,
    other_stuff = 1 << 1,
    ... more variants
};
```

But why are we even using `uint8_t`? That suggests there is some very good reason these integers need to be exactly 8 bits, but... is there? I would just use `int`, or `unsigned` if you are so inclined. Then you don't even need that integer type clause on the enum.

configdialog.cpp
```diff
@@ -518,6 +535,271 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		colors_incorrect_length = 1 << 4,
+		colors_invalid_characters = 1 << 5,
+	};
+
+	vector<wstring> names;
+	vector<vector<wstring>> colors;
+	vector<uint8_t> is_valid;
+
+	for (; palettes->find_first_of(valid_chars, current_index) != wstring::npos && palettes->find_first_of(L':', current_index) != wstring::npos; palettes_amount++) {
+		graceful_exit = (palettes->find_first_of(L'~', current_index) != wstring::npos);
```
pierce-smith1: Again, I'm not going to comb over this line-by-line, because it hurts me. It just needs to work for now - we will determine that by testing it at runtime, once everything is done.

However, if you do want me to go through specific examples of things I don't like about this and how I would make it more understandable, just let me know. We can go through some of it when I have more mental bandwidth to do so.

configdialog.cpp
```diff
@@ -518,6 +535,271 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		int num = MessageBox(
+			NULL,
+			import_confirmation.c_str(),
+			L"Palette Customizer",
+			MB_YESNO | MB_ICONINFORMATION
+		);
+		import = (num == IDYES);
+	}
+
+	if (import) {
```
pierce-smith1: Please for the love of God make this a separate function. One function for validating, one for actually importing. This can't possibly get any longer.

configdialog.cpp
pierce-smith1: Because our memory management needs are simple, we are quite alright using raw pointers and managing them ourselves. The smart pointer here is maybe better in a vacuum, but the inconsistency with the rest of the code is confusing (why is _this_ a smart pointer? Why is it special?)

pierce-smith1 (changes requested):
Oh dear.

(I instinctively clicked on "request changes" because the code makes me upset, but as you'll read I'm not really asking for any major changes right now.)

configdialog.cpp
```diff
@@ -518,6 +535,271 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		);
+		import = (num == IDYES);
+	}
+	else if (count_invalid > 0) {
+		wstring error_info = L"One or more palettes have errors:\n";
+		for (size_t i = 0, invalid = 0; i < palettes_amount; i++) {
+			if (is_valid[i] != static_cast<uint8_t>(Errors::none)) {
+				invalid++;
+				wstring name = names[i];
+				while (name.find_first_of(L'\r') != wstring::npos) {
```
pierce-smith1: This is something I do want to make sure is changed though. We cannot have newlines, they are too dangerous.

configdialog.cpp
McAwesome123: originally it was for extra validation, but I've decided that it's completely unnecessary.

configdialog.cpp
```diff
@@ -578,6 +598,296 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		}
+	}
+
+	if (!import_string_had_terminator && invalid_palette_count <= 1 && (invalid_palette_count == 0 || *palettes_has_errors.rbegin() != error_none)) {
+		std::wstring error_string = L"No terminator character (~) found at the end of the input. This may be caused by the inputted text being too long.";
+		auto palette_name = palettes_name.rbegin();
+		if (*palettes_has_errors.rbegin() != error_none) {
+			palette_name++;
+		}
+		error_string.append(L"\nLast valid palette found: " + *palette_name);
```
pierce-smith1: If `palettes_name` only has one element and the previous if statement triggers, then `palette_name` is an invalid iterator that gets dereferenced, causing a crash in debug. You can replicate this by pasting anything that ends with a colon into the palette box (or funnily enough, just typing a single `:` in there)

configdialog.h
```diff
@@ -81,6 +83,41 @@ class PaletteCustomizeDialog {
...
+	std::wstring get_valid_palette_names(size_t max_length);
+	void import_palettes();
+
+private:
+	std::vector<std::wstring> palettes_name;
+	std::vector<std::vector<std::wstring>> palettes_raw_colors;
+	std::vector<unsigned int> palettes_has_errors;
+	bool import_string_had_terminator;
+
+	const static inline std::wstring valid_chars = L"qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM0123456789-_!?:";	// Note: Spaces are valid characters but not included in this string
```
pierce-smith1: It's sort of weird that `(` and `)` are not valid characters even though a duplicate palette import creates a palette name with for example `(2)` in it. This means it is possible to organically create a palette that can never be imported. I think that parens can be added to the valid characters list.

configdialog.cpp
```diff
@@ -763,6 +1073,99 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
...
+			switch (LOWORD(wparam)) {
+				case IDOK: {
+					std::wstring palettes = palettes_buffer;
+					PalettesImport *imported_palettes = PalettesImport::parse_palettes_string(palettes);
+					delete[] palettes_buffer;
+
+					if (imported_palettes->get_error_string(1).has_value()) {
+						int import_anyway = MessageBox(
+							NULL,
+							(L"The inputted text contains one or more problems. Are you sure you want to import anyway?\n\nThe following "
```
pierce-smith1, line 1166: The following import string:
```
aemil: #56eb8e;#84f5c3;#1d9550;#e29a56;#e16a72;#dff9eb;#966336;
ascent: #1963c4;#1963c4;#0b407d;#8c96dd;#8c96dd;#ffffff;#506bbc;
autumn: ;
bliss: #73981e;#73981e;#3d5317;#6a96f2;#282438;#eaf2ff;#3b73ee;
~
```
Results in this message box:
image
It says it will import autumn, but it (correctly) only imports aemil, ascent, and bliss.

configdialog.cpp
```diff
@@ -320,6 +320,26 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
...
+				(DLGPROC) ScreenSaverImportExportPalettesDialog,
+				(LPARAM) exported_palettes
+			);
+
+			if (imported_palettes == nullptr) {
+				break;
+			}
+
+			imported_palettes->import_palettes();
+			refresh_palette_list();
```
pierce-smith1, line 339: After the palettes are imported, we should select one of them (maybe the first or last one imported). Other pieces of code expect a palette to always be selected if possible - try deleting all your palettes, then importing some, then clicking on the empty space in the palette list, you'll get an invalid dereference and a crash in debug.

I think all you have to do is set `m_current_palette` to something before calling `refresh_palette_list`. I suggest you make `import_palettes()` return a list of the palette definitions it actually imported and pull something from there.

configdialog.cpp
```diff
@@ -578,6 +598,296 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
+		count++;
+	}
+
+	palette_names.pop_back();
+	palette_names.pop_back();
+
+	return palette_names;
+}
+
+void PalettesImport::import_palettes() {
```
pierce-smith1: If you import a decent number of palettes (say, around 40), it's very slow. I think something is just generally up with the palette dialog though, as _everything_ gets very slow at around 30-40 palettes, and that should absolutely not be the case. I must be doing something very stupid somewhere.

pierce-smith1 (changes requested):
This was everything I could find after a good 45 minutes toying with it.

configdialog.cpp
```diff
@@ -578,6 +598,296 @@ void PaletteCustomizeDialog::get_and_save_color(int palette_index) {
...
 
+std::wstring PaletteCustomizeDialog::export_palettes() {
+	auto all_palettes = m_palette_repo.get_all_custom_palettes();
+	std::wstring palettes_string = L"";
+
+	for (const auto &palette : all_palettes) {
+		auto name = palette.name;
+		auto *data = palette.data;
+
+		palettes_string.append(palette.name + L": " + m_palette_repo.serialize(*data) + L";\r\n");
```
pierce-smith1: It took me way too long to realize this, but it's actually possible to create a palette with a colon in the name - I made it a valid character when creating palettes. The import obviously isn't gonna work right with colons in the palette names. There are two choices for resolving this:

1) Use a different separator than a colon here (maybe an equals sign?)
2) Just assume that nobody has made a palette with a colon in it, silently remove the colon from the list of valid characters when making a palette, and trust that nothing will go wrong? It sounds bad but I think this is a relatively safe bet.

Note that I think we should NOT choose option 3, which is to check for a `:` in the exporter and/or the importer and remove it or replace it or warn about it or something. This is complicated enough. We need to be stupider.

configdialog.cpp
pierce-smith1: [Something was up.](https://github.com/pierce-smith1/yok.scr/pull/59)

configdialog.h
McAwesome123: That is... a bit of an oversight. But yeah, we can probably add the parentheses to the valid characters.

configdialog.cpp
McAwesome123: clearly, I haven't learned from the last time I had a crash like this.
I can see why I might've missed it and it's because I keep testing this thing in release.

configdialog.cpp
McAwesome123: okay, this was just a complete brain fart on my end.
The `get_valid_palette_names` function apparently just returns every single name and I didn't even attempt to make it check whether or not it's valid.
I have no clue how I missed this.

configdialog.cpp
McAwesome123: oh boy...

configdialog.cpp
McAwesome123: :lkconcern:

configdialog.cpp
McAwesome123: _Somehow_ I just never noticed that colons are valid characters. I think I'll go with option 1 even though it looks a bit worse imo.

pierce-smith1 (approved):
Good enough - I think we're done playing catch with this. It will need to be cleaned up, but we'll deal with that later.

Issue #50, opened 2024-03-28
Add a .png export for custom palettes


pierce-smith1:
This was suggested to me directly. It's a pretty weird feature from the screensaver's perspective, but I really like it. I guess the idea is that after you design a palette, you can export a .png for every bitmap with that palette applied, turning yok.scr from a screensaver into a bona fide yokin factory.

This sounds pretty hard, and potentially not even practically possible depending on Windows limitations. This article looks extremely promising though: https://learn.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-converting-a-bmp-image-to-a-png-image-use

PR #57, opened 2024-03-31
PNG exporting for palettes

pierce-smith1:
#50 

This PR adds a "Save palette as images" option to the palette customizer dialog, allowing you to export a set of fully rendered yokin PNGs in the selected palette.

### Concerns

* We're depending on some totally new stuff to do this (GDI+ and the Vista-era file picker dialog), so there's a lot of totally new and exciting ways for shit to go wrong that we will only discover when this is in other people's hands.
* I don't really like the UX of the saving. When you're picking where to save the images, the dialog is in folder picker mode, so it doesn't show any existing files in the folders as you're browsing through them. So you might be browsing through stuff and open a folder and think "Oh great, there's nothing in here, good place the dump the yokins!" but actually there's loads of shit in there and you just made a mess. Also - dumping 20+ PNGs into a folder is just messy and feels impolite - people might not expect it.
* I wrote this while very emotionally exhausted, so there's probably something glaringly wrong with it.


configdialog.cpp
```diff
@@ -488,6 +529,79 @@ void PaletteCustomizeDialog::delete_current_palette() {
...
+			PixelFormat32bppARGB,
+			texture_data
+		);
+
+		// This is kind of gross, but the alternative is a 20-something line function
+		// with calls to GetImageEncoders and other crap... just to get the ID of the
+		// PNG encoder... seriously?
+		// I have no good reason to believe this ID will be different between machines
+		// or that any given machine will not have this encoder. So, blah.
+		struct __declspec(uuid("{557cf406-1a04-11d3-9a73-0000f81ef32e}")) PngEncoder;
```
McAwesome123, line 684: What could go wrong‽

resourcew.h
```diff
@@ -98,6 +98,8 @@
 #define IDC_PALDLG_INSERT_PREDEFINED    1060
 #define IDC_PALDLG_CHOOSE_PREDEFINED    1061
 #define IDC_PALDLG_PREDEFINED_PREVIEW   1062
+#define IDC_PALDLG_EXPORT               1063
+#define IDC_PALDLG_PNG_EXPORT           1063
```
McAwesome123: a bit of a nitpick, you have two defines for the same thing

McAwesome123 (commented):
So, from what I've gathered, pretty much your only option for having a folder picker where users can still see files is to make your own.
An alternative to that would be to add one of those message boxes that you hate to notify the user that the folder isn't empty.

Also worth noting: the program will (without asking!) overwrite any images with the same name as the ones being saved. Since the resulting images are named the same as the base bitmaps, this means that trying to save multiple palettes to the same folder will overwrite the previous export.

...I am a bit on the fence about approving this one.

configdialog.cpp
```diff
@@ -488,6 +529,79 @@ void PaletteCustomizeDialog::delete_current_palette() {
...
+		// relative to how we store the pixels.
+		// So, flip the image across the y.
+		for (int y = 0; y < BITMAP_WH / 2; y++) {
+			int y_prime = BITMAP_WH - 1 - y;
+
+			for (int x = 0; x < BITMAP_WH; x++) {
+				int i = x + y * BITMAP_WH;
+				int i_prime = x + y_prime * BITMAP_WH;
+
+				auto red = texture_data[i * 4 + RED];
```
pierce-smith1: Note to self, this could probably be dramatically simplified by using std::swap instead

configdialog.cpp
```diff
@@ -529,12 +531,17 @@ void PaletteCustomizeDialog::delete_current_palette() {
 	}
 }
 
+std::wstring PaletteCustomizeDialog::get_png_export_path(const std::wstring &base_path, const std::wstring &palette_name) {
+	return std::format(L"{}\\yokins-{}-{}", base_path, palette_name, rand());
```
pierce-smith1: Suffixing the folder with a random number is a super simple way to ensure that exporting a palette with the same name as one you've already exported won't cause any overwriting (or at least that it will be highly unlikely).

McAwesome123: So you're saying there's a chance...

McAwesome123: Well, I found another oversight.
This isn't given a variable seed, so it gives the exact same sequence of numbers after every restart.

McAwesome123 (approved):
:lkapprove:

Issue #58, opened 2024-04-05
Having ~40 palettes causes the palette customizer to slow to a crawl


pierce-smith1:
This shouldn't be happening.

PR #59, opened 2024-04-05
Make palette customizer dialog not slow as shit

pierce-smith1:
#58 

It turns out the (primary) cause of this was converting color data from the registry from a wide string to a short string every time the dialog refreshed - the function Windows uses to do this (`WideCharToMultiByte`) is insultingly slow. I just made the palette data deserializer work on wide strings directly, which involved some other code changes, but not too bad.

I also tried to solve the palette dialog flickering wildly when you scroll through the palettes, which is due to invalidating the _entire_ dialog window when the only thing that needs to be repainted is the color buttons. I didn't really like what I had to do to get that to work... but it seems alright?


configdialog.cpp
```diff
@@ -381,7 +381,36 @@ void PaletteCustomizeDialog::refresh() {
...
+
+	// In Windows Vista+, calling GetWindowRect on a dialog does not 
+	// just return the bounds of the dialog;
+	// it gives the bounds of the dialog _including the drop shadow!_
+	// There is a function for getting the size without the drop shadow,
+	// but it does NOT respect DPI settings, so it's effectively useless.
+	// (and it requires linking against dwmapi... a whole dependency just for this!?)
+	// Fuck it! I measured roughly how big the drop shadow is on my
+	// machine and offset the rectangle by that.
+	// Why do you hate me, billy?
```
McAwesome123, line 395: :lkanim:

McAwesome123 (approved):

Issue #49, opened 2024-03-27
Add some way of previewing the rest of the bitmaps


McAwesome123:
This could be useful for previewing how the rest of the yokers and others look like without having to just run the screensaver.

PR #62, opened 2024-04-06
Preview all bitmaps

pierce-smith1:
#49 

image

This PR adds some buttons under the preview in the palette customizer to let you change the bitmap you're previewing. I also added a warning when the current bitmap doesn't make use of every color in the palette, since I can see people trying to edit the colors on a bitmap and being confused when they don't change because they aren't there in the first place.


resourcew.h
```diff
@@ -99,6 +99,10 @@
 #define IDC_PALDLG_CHOOSE_PREDEFINED    1061
 #define IDC_PALDLG_PREDEFINED_PREVIEW   1062
 #define IDC_PALDLG_COLOR_BUTTONS_REGION 1063
+#define IDC_PALDLG_PREV_BITMAP          1064
+#define IDC_BUTTON2                     1065
+#define IDC_PALDLG_NEXT_BITMAP          1065
```
McAwesome123, line 103: Nitpick: `IDC_BUTTON2` is unused

configdialog.cpp
```diff
@@ -357,6 +381,19 @@ HBRUSH PaletteCustomizeDialog::handle_color_button_message(WPARAM wparam, LPARAM
 }
 
 void PaletteCustomizeDialog::refresh() {
+	HWND not_all_colors_warning = GetDlgItem(m_dialog, IDC_PALDLG_NOT_ALL_COLORS_WARNING);
+	const static std::set<Bitmaps::Definition> bitmaps_with_all_colors = {
+		Bitmaps::Lksix,
+		Bitmaps::Lkhusk,
```
McAwesome123: Lkhusk lacks an eye color

McAwesome123 (changes requested):
Bit of an oversight with `bitmaps_with_all_colors` but otherwise good

McAwesome123 (approved):

PR #69, opened 2024-04-07
Various palette customizer fixes

### LINKED ISSUE #53: When deleting palettes in the palette customizer, the palette your selection is moved to is effectively random
This is debatably not a bug, but it's odd behavior that I'm almost 100% certain will get someone in trouble.

Say you have palettes like this, and you have the last one selected:

```
a
b
c
d
e <--
```

Then let's say you want to delete everything except for `a` and `b`. Intuitively, you may want to just spam the delete button 3 times, thinking that your selection will move to `d` then to `c`. But no, in reality it's based on the order you added them, and to make things worse the order they are displayed is _different from that_ (it's alphabetical), so it's effectively random. Your selection may suddenly jump to `a` or `b` after deleting `e` for instance.

The fix for this makes the selection move to whatever item was before the one you deleted, and if you deleted the first one, then it becomes whatever the new first item is. That may involve making the display order not alphabetical.

### LINKED ISSUE #60: When editing a color in a palette, the color dialog does not start with that color selected
Title. Pretty annoying.

It would be nice if it also filled in the hex code in the paste box, but honestly that is probably more effort than it's worth.

### LINKED ISSUE #66: Import dialog can add duplicate palettes
The name check in the palette importer is case-sensitive, while Windows registry keys are case-insensitive.
This can lead to having multiple references to the same palette which can really fuck things up.

### LINKED ISSUE #68: Have a check for if a palette in the index actually exists
There should probably be a check for this so that when the index contains a palette that doesn't actually exist, the screensaver handles it properly instead of doing Weird Shit™.

Things I've seen happen so far as a result of this range from crashes on debug to Windows Defender popups(!) on release.

McAwesome123:
This pull request is meant to address various, (mostly) minor issues with the palette customizer.

Changes include:
- Palettes are no longer sorted in alphabetical order (instead they appear in the order they were added)
- Deleting a palette automatically selects the palette above (or below) ([#53](https://github.com/pierce-smith1/yok.scr/issues/53))
- Color hex code now gets filled in and automatically updates ([#60](https://github.com/pierce-smith1/yok.scr/issues/60))
- Import dialog no longer creates duplicate palette references due to case-insensitive name-checking ([#66](https://github.com/pierce-smith1/yok.scr/issues/66))
- Undefined palettes are automatically removed from the index ([#68](https://github.com/pierce-smith1/yok.scr/issues/68))


configdialog.cpp
```diff
@@ -614,13 +614,23 @@ void PaletteCustomizeDialog::delete_current_palette() {
 		return;
 	}
 
+	auto all_palettes = m_palette_repo.get_all_custom_palettes();
+	size_t index = 0;
+	if (all_palettes.size() > 1) {
+		for (size_t i = 0; i < all_palettes.size(); i++) {
+			if (all_palettes[i].name == m_current_palette->name) {
+				index = (i > 0 ? i - 1 : i + 1);
```
McAwesome123: I doubt this is the most elegant way to do this, but it's the first thing I thought of.

configdialog.cpp
```diff
@@ -1256,9 +1268,66 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
...
+				case COLOR_RED:
+				case COLOR_GREEN:
+				case COLOR_BLUE: {
+					switch (HIWORD(wparam)) {
+						case EN_CHANGE: {
+							if (ignore_hex_change > 0) {
+								ignore_hex_change--;
+								return FALSE;
+							}
+							ignore_rgb_change = 1;
```
McAwesome123, line 1280: :fnyoy:

configdialog.cpp
```diff
@@ -1256,9 +1268,66 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
...
+								);
+							});
+
+							if (red.empty() || green.empty() || blue.empty()) {
+								return FALSE;
+							}
+
+							std::wstring hex_code = std::format(L"#{:02x}{:02x}{:02x}", std::stoi(red), std::stoi(green), std::stoi(blue));
+							HWND hex_input = GetDlgItem(dialog, IDC_COLORDLG_HEX_CODE);
+							Edit_SetText(
```
McAwesome123, line 1320: Despite the madness that is... the above code, so far, this has worked through everything I've thrown at it.

McAwesome123:
Depending on how I feel, I might potentially lump [#68](https://github.com/pierce-smith1/yok.scr/issues/68) into this.

configdialog.cpp
```diff
@@ -1256,9 +1268,66 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
...
 			switch (LOWORD(wparam)) {
+				case COLOR_RED:
+				case COLOR_GREEN:
+				case COLOR_BLUE: {
+					switch (HIWORD(wparam)) {
+						case EN_CHANGE: {
+							if (ignore_hex_change > 0) {
+								ignore_hex_change--;
+								return FALSE;
+							}
```
McAwesome123, line 1285: :fnyoy:

McAwesome123:
Depending on how I feel, I might potentially lump [#68](https://github.com/pierce-smith1/yok.scr/issues/68) into this.

configdialog.cpp
```diff
@@ -614,13 +614,23 @@ void PaletteCustomizeDialog::delete_current_palette() {
 		return;
 	}
 
+	auto all_palettes = m_palette_repo.get_all_custom_palettes();
+	size_t index = 0;
```
pierce-smith1: A more descriptive name would be nice. Maybe `next_index_to_select` or something.

configdialog.cpp
```diff
@@ -1235,6 +1245,8 @@ LRESULT CALLBACK CustomColorDialog(HWND dialog, UINT message, WPARAM wparam, LPA
 	};
 
 	static UINT set_rgb_message_code = RegisterWindowMessage(SETRGBSTRING);
+	static int ignore_rgb_change = 0;
+	static int ignore_hex_change = 0;
```
pierce-smith1: WTF are these and why?

configdialog.cpp
pierce-smith1: It looks fine, but holy moly that indentation is packing up to move across the ocean. We can fix that later though.

pierce-smith1 (approved):
I love these fixes, makes everything so much smoother and less scary.

configdialog.cpp
McAwesome123: Using `Edit_SetText` on the hex code or RGB values will immediately run the `EN_CHANGE` code for them.
In general, that tends to quickly cause the program to shit itself with a stack overflow exception, so these exist to prevent that.

It might be hard to tell, but I _hate_ this solution.

configdialog.cpp
McAwesome123: I don't... disagree.

configdialog.cpp
pierce-smith1: Wow fuck that. A little bit of googling seems to confirm that you can't avoid that EN_CHANGE notification sent via SetText at all, and that some kind of check like this is how you're "supposed" to do it.

The names of these might be more descriptive as `updates_to_ignore` or something. I was very confused why they were integers at first. This is also a great time for a comment to explain why the hell you have to do this.

palettes.cpp
```diff
@@ -247,6 +247,14 @@ std::optional<Palettes::Definition> PaletteRepository::get_palette(const std::ws
 std::vector<Palettes::Definition> PaletteRepository::get_all_custom_palettes() {
 	auto items = m_map.items();
 
+	// Remove any undefined palette indices
+	for (auto item = items.begin(); item != items.end(); item++) {
+		if (!get_palette(item->first).has_value()) {
+			remove_palette(item->first);
+			item = items.erase(item) - 1;
```
McAwesome123, line 254: I wanted to have this in `get_palette` but I couldn't figure out a simple way to make `std::transform()` ignore an entry.
But as this function always seems to run when doing anything with custom palettes, it should work just as well (or maybe better?)

McAwesome123 (commented):
Besides minor changes like variable names, this also addresses #68.
Primarily because it only took about half an hour (or less) to do so.

pierce-smith1 (approved):
🆗 🆒

PR #70, opened 2024-04-10
Make palette name length checking more gooder

### LINKED ISSUE #67: Add palette option can exceed maximum name length when a name is already taken
When a number is being appended to an already existing palette name, there is no check for how long the name is, meaning it can end up being longer than 63 characters.
While this doesn't seem to break most other functionality, it would not be possible to import such a palette.

pierce-smith1:
#67

This also tries to abstract some complexity by making both the new palette dialog and the import dialog use the same code for de-duplicating names.


Resource.rc
```diff
@@ -126,7 +126,7 @@ FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
     DEFPUSHBUTTON   "OK",IDOK,256,293,50,14
     GROUPBOX        "Palettes",IDC_STATIC,7,7,143,300
-    LISTBOX         IDC_PALDLG_PALETTE_LIST,13,33,130,212,LBS_NOINTEGRALHEIGHT | WS_VSCROLL | WS_TABSTOP
+    LISTBOX         IDC_PALDLG_PALETTE_LIST,13,33,130,212,LBS_SORT | LBS_NOINTEGRALHEIGHT | WS_VSCROLL | WS_TABSTOP
```
McAwesome123: Why's this being changed back?

McAwesome123 (approved):
Alright, cool.

Resource.rc
pierce-smith1: What?! I didn't ask for this!

PR #72, opened 2024-06-12
Braver Llokies

McAwesome123:
Ever since "fixing" bravery some time ago, I've been disappointed at how little of an effect it had, save for setting it to extremely low values.

This PR aims to change that.

The yockies now have more of a tendency to roam around near their home.

### DIFF
--- a/configdialog.cpp
+++ b/configdialog.cpp
@@ -62,10 +62,6 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 				m_dialog,
 				(DLGPROC) ScreenSaverPaletteCustomizeDialog
 			);
-			for (HBRUSH brush : PaletteCustomizeDialog::used_brushes) {
-				DeleteObject(brush);
-			}
-			PaletteCustomizeDialog::used_brushes.clear();
 			break;
 		}
 		case IDC_YONK_PATTERN: {
@@ -446,8 +442,6 @@ HBRUSH PaletteCustomizeDialog::handle_color_button_message(WPARAM wparam, LPARAM
 		std::get<GREEN>(color),
 		std::get<BLUE>(color)
 	));
-
-	used_brushes.push_back(brush);
 
 	// This may be a resource leak because brushes are supposed to be freed
 	// after their use, but I have no idea how to get a reference to this brush
--- a/configdialog.cpp
+++ b/configdialog.cpp
@@ -62,6 +62,10 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 				m_dialog,
 				(DLGPROC) ScreenSaverPaletteCustomizeDialog
 			);
+			for (HBRUSH brush : PaletteCustomizeDialog::used_brushes) {
+				DeleteObject(brush);
+			}
+			PaletteCustomizeDialog::used_brushes.clear();
 			break;
 		}
 		case IDC_YONK_PATTERN: {
@@ -108,12 +112,19 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 BOOL ConfigDialog::slider_changed(WPARAM wparam, HWND slider) {
 	auto value = SendMessage(slider, TBM_GETPOS, 0, 0);
 
-	auto opt = *std::find_if(Cfg::All.begin(), Cfg::All.end(), [&](const auto &opt) {
+	auto &opt = *std::find_if(Cfg::All.begin(), Cfg::All.end(), [&](const auto &opt) {
 		return opt.dialog_control_id == GetDlgCtrlID(slider);
 	});
 
 	m_current_config[opt] = decodef(value);
 
+	return FALSE;
+}
+
+BOOL ConfigDialog::checkbox_checked(WPARAM wparam, HWND checkbox, const Cfg::Definition &option) {
+	bool checked = Button_GetCheck(checkbox) == BST_CHECKED;
+	m_current_config[option] = checked ? 1.0 : 0.0;
+	refresh();
 	return FALSE;
 }
 
@@ -131,13 +142,6 @@ BOOL ConfigDialog::combobox_changed(HWND combobox, int option_type) {
 			break;
 	}
 
-	return FALSE;
-}
-
-BOOL ConfigDialog::checkbox_checked(WPARAM wparam, HWND checkbox, const Cfg::Definition &option) {
-	bool checked = Button_GetCheck(checkbox) == BST_CHECKED;
-	m_current_config[option] = checked ? 1.0 : 0.0;
-	refresh();
 	return FALSE;
 }
 
@@ -250,7 +254,8 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 			auto current_bitmap_position = Bitmaps::All.find(m_current_preview_bitmap);
 
 			auto bitmaps_start = Bitmaps::All.begin();
-			auto bitmaps_last = --Bitmaps::All.end();
+			auto bitmaps_last = Bitmaps::All.end();
+			bitmaps_last--;
 
 			if (control_id == IDC_PALDLG_PREV_BITMAP && current_bitmap_position == bitmaps_start) {
 				m_current_preview_bitmap = *bitmaps_last;
@@ -301,7 +306,7 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 				break;
 			}
 
-			auto new_data = LOWORD(wparam) == IDC_PALDLG_DUPE_PALETTE
+			auto &new_data = LOWORD(wparam) == IDC_PALDLG_DUPE_PALETTE
 				? m_current_palette->data
 				: *DefaultPalette.data;
 
@@ -329,7 +334,7 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 				return FALSE;
 			}
 
-			auto palette = *std::find_if(Palettes::All.begin(), Palettes::All.end(), [&](const Palettes::Definition &palette) {
+			auto &palette = *std::find_if(Palettes::All.begin(), Palettes::All.end(), [&](const Palettes::Definition &palette) {
 				return *predefined_palette_name == palette.name;
 			});
 
@@ -351,12 +356,16 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 		case IDC_PALDLG_PNG_EXPORT: {
 			if (m_current_palette) {
 				IFileDialog *file_dialog;
-				CoCreateInstance(
+				HRESULT hr = CoCreateInstance(
 					CLSID_FileOpenDialog,
 					NULL,
 					CLSCTX_INPROC_SERVER,
 					IID_PPV_ARGS(&file_dialog)
 				);
+
+				if (hr != S_OK) {
+					break;
+				}
 
 				DWORD flags;
 				file_dialog->GetOptions(&flags);
@@ -385,8 +394,8 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 
 				selected_folder->Release();
 			}
-      break;
-    }
+		  break;
+		}
 		case IDC_PALDLG_IMPORT_EXPORT_PALETTES: {
 			std::wstring *exported_palettes = new std::wstring(export_palettes());
 			PalettesImport *imported_palettes = (PalettesImport *) DialogBoxParam(
@@ -437,6 +446,8 @@ HBRUSH PaletteCustomizeDialog::handle_color_button_message(WPARAM wparam, LPARAM
 		std::get<GREEN>(color),
 		std::get<BLUE>(color)
 	));
+
+	used_brushes.push_back(brush);
 
 	// This may be a resource leak because brushes are supposed to be freed
 	// after their use, but I have no idea how to get a reference to this brush
@@ -684,7 +695,7 @@ void PaletteCustomizeDialog::apply_palette_to_preview(HWND dialog, HANDLE previe
 	HDC memory_context = CreateCompatibleDC(GetDC(dialog));
 
 	SelectBitmap(memory_context, preview_bitmap);
-	RGBQUAD colors[_PALETTE_SIZE];
+	RGBQUAD colors[_PALETTE_SIZE] { };
 
 	// I REALLY don't want to go down a rabbit hole of figuring out how to make this bitmap
 	// actually transparent (it's an absolute miracle any of this works _at all_!), so we
@@ -775,10 +786,10 @@ std::wstring PaletteCustomizeDialog::export_palettes() {
 	std::wstring palettes_string = L"";
 
 	for (const auto &palette : all_palettes) {
-		auto name = palette.name;
+		auto &name = palette.name;
 		auto *data = palette.data;
 
-		palettes_string.append(palette.name + L" " + PalettesImport::palette_name_end + L" " + m_palette_repo.serialize(*data) + PalettesImport::palette_color_end + L"\r\n");
+		palettes_string.append(name + L" " + PalettesImport::palette_name_end + L" " + m_palette_repo.serialize(*data) + PalettesImport::palette_color_end + L"\r\n");
 	}
 	palettes_string.append(PalettesImport::input_terminator);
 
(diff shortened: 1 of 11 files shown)


noise.cpp
```diff
@@ -124,14 +124,20 @@ double PerlinNoise::interpolate(double a, double b, double w) {
 	return (b - a) * (3.0 - w * 2.0) * w * w + a;
 }
 
-double Noise::wiggle(double base, double min, double max, double step) {
-	bool up = random() < 0.5;
+// rand_exp: Bias for the random number generator. >1 outputs higher numbers on average, <1 outputs lower numbers on average
+// distance_exp: Bias for the distance from the center. >1 increases the percieved distance, <1 decreases the percieved distance
+double Noise::wiggle(double base, double center, double min, double max, double step, double rand_exp, double distance_exp) {
```
pierce-smith1: This function has a _lot_ of parameters. This can be a problem because it makes it hard to call the function and/or read the call sites since I have to remember what all 7 of these things do. in particular when reading a call site you just see 7 random numbers with no context attached to them and it can get really confusing.

Personally, I would restore Noise::wiggle to the way it was, then introduce this as a function in the Yonker class with some of these parameters just hardcoded as const doubles like you do in the body of update(). It might seem counter-intuitive to switch from parameters to hardcoding but this is such a specialized function already that I don't think it is useful in a highly generic form like this.

sprite.cpp
```diff
@@ -9,7 +9,7 @@
 using std::get;
 
 Sprite::Sprite(const Texture *texture, const Point &home, const bool has_trail)
-	: m_texture(texture), m_home(home), m_relpos(0.0, 0.0), m_size(cfg[Cfg::SpriteSize] / 1000.0f), m_trail_start_index(0)
+	: m_texture(texture), m_home(home), m_relpos(0.0, 0.0), m_relpos_tendency(0.0, 0.0), m_size(cfg[Cfg::SpriteSize] / 1000.0f), m_trail_start_index(0)
```
pierce-smith1: Nitpick, this member initializer list is getting long enough it should probably put each member on its own line.

sprite.cpp
```diff
@@ -115,18 +115,53 @@ void Yonker::update(Context &ctx) {
 	// In little steps up and down they'll roam,
 	// But never too far outside their home.
 	if (cfg[Cfg::HomeDrift] >= 0.000001) {
+		// currently, these exist for ease of modifying the values
+		const unsigned int time_between_tendency_changes = 60 * 30;
```
pierce-smith1: Nitpick, when naming a variable like this it really helps to include the units (what does "time" mean? Milliseconds? Seconds?). In this case the unit is frames, so consider something like `frames_between_tendency_changes`.

sprite.cpp
```diff
@@ -115,18 +115,53 @@ void Yonker::update(Context &ctx) {
...
+		const double shake_divisor = 2.5;
+		const double tendency_randomness_exponent = 1.5;
+		const double tendency_distance_exponent = 1;	// doesn't do anything
```
pierce-smith1: Then why is it there?

pierce-smith1 (changes requested):
This is cool and I love it, but it's revealing some big problems with how yokies that roam really far from their home interact with certain patterns. Try turning bravery and shakiness all the way up on Lissajous, Rose, and Bouncy, it turns into a fucking fruit blender really quickly. 

The core issue seems to be that those three patterns simply don't expect sprites to ever wrap, but if their final position drifts far enough away from home, they will, and it causes weird shit to happen. This problem definitely wasn't added by this PR but we should at least think about how to address it. My brain is fried like an egg today though so I don't have anything for you right now.

McAwesome123:
> This is cool and I love it, but it's revealing some big problems with how yokies that roam really far from their home interact with certain patterns. Try turning bravery and shakiness all the way up on Lissajous, Rose, and Bouncy, it turns into a fucking fruit blender really quickly.
> 
> The core issue seems to be that those three patterns simply don't expect sprites to ever wrap, but if their final position drifts far enough away from home, they will, and it causes weird shit to happen. This problem definitely wasn't added by this PR but we should at least think about how to address it. My brain is fried like an egg today though so I don't have anything for you right now.

maybe i should've tested this with more patterns besides lattice

sprite.cpp
McAwesome123: not really any reason other than for the sake of having everything

noise.cpp
McAwesome123: I'm not entirely sure what you're asking me to do here.

sprite.cpp
pierce-smith1: I thought that everywhere this function was called you were using the same last two parameters and I was basically telling you to factor those parameters out into constants, but after looking at the code for more than 2 seconds that obviously isn't the case. Let's talk about it later, I think we can come up with a better solution.

noise.cpp
pierce-smith1: I thought that everywhere this function was called you were using the same last two parameters and I was basically telling you to factor those parameters out into constants, but after looking at the code for more than 2 seconds that obviously isn't the case. Let's talk about it later, I think we can come up with a better solution.

configdialog.cpp
```diff
@@ -62,6 +62,10 @@ BOOL ConfigDialog::command(WPARAM wparam, LPARAM lparam) {
 				m_dialog,
 				(DLGPROC) ScreenSaverPaletteCustomizeDialog
 			);
+			for (HBRUSH brush : PaletteCustomizeDialog::used_brushes) {
```
pierce-smith1: I would just completely avoid trying to delete these brushes. The bottom line is that there is no evidence at all to suggest they are causing any sort of problem, and complexity cost involved with having even more meaningless code and tracking even more state _just_ to try to delete them is not even close to worth it in my eyes.

configdialog.cpp
```diff
@@ -250,7 +254,8 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 			auto current_bitmap_position = Bitmaps::All.find(m_current_preview_bitmap);
 
 			auto bitmaps_start = Bitmaps::All.begin();
-			auto bitmaps_last = --Bitmaps::All.end();
+			auto bitmaps_last = Bitmaps::All.end();
+			bitmaps_last--;
```
pierce-smith1, line 254: Why?

configdialog.cpp
```diff
@@ -351,13 +356,17 @@ BOOL PaletteCustomizeDialog::command(WPARAM wparam, LPARAM lparam) {
 		case IDC_PALDLG_PNG_EXPORT: {
 			if (m_current_palette) {
 				IFileDialog *file_dialog;
-				CoCreateInstance(
+				HRESULT hr = CoCreateInstance(
```
pierce-smith1, line 359: Why?

configdialog.cpp
```diff
@@ -803,7 +814,7 @@ std::wstring PaletteCustomizeDialog::get_unique_suffixed_name(const std::wstring
 		return palette.name;
 	});
 
-	auto new_name = base;
+	std::wstring new_name = base;
```
pierce-smith1, line 811: Why?

sprite.h
```diff
@@ -29,20 +29,41 @@ class Sprite : public Identifiable<Empty> {
...
+
+	template<typename T>
+	double randomize_tendency_variable(pair_randomness<T> tendency_pair);
+
+	inline const static pair_randomness<size_t> default_frames_between_tendency_changes = { 10 * 30, 0.5 };
+	inline const static pair_randomness<double> default_tendency_distance_from_home_bias = { 1.5, 0.2 };
+	inline const static pair_randomness<double> default_wiggle_amount_bias = { 1.0, 0.25 };
+	inline const static pair_randomness<double> default_distance_from_tendency_bias = { 2.5, 1.0 };
+	inline const static double shake_divisor = 1 / 0.3;
+	inline const static double home_drift_divisor = 10.0;
```
pierce-smith1, line 66: I don't have any problem with doing some complicated math to make more interesting randomness, but this complexity is bleeding all over the place. This sprite class now looks waaaay more complicated than it really is because there are a dozen or more members flying all over the fucking place (just look at the constructor of this class, my god.) If we're going to make wiggling and tendency so complicated, then let's stuff all that complexity into a box that we can ignore, because otherwise in the future we're going to come back to this class and it's going to totally overwhelm us. A sprite is really simple, we shouldn't be overwhelmed.

I think the best approach here is to make a SpriteWiggler class that becomes a single member of Sprite which is itself responsible for all this crap:
* all five of the instance variables you added just below here
* all these statics above, plus the templates
* the `randomize_tendency` and related functions
 
Then, in our `Sprite::update` function, we can just call `m_home = m_sprite_wiggler.wiggle(m_home)` or something like that and be fucking done with it.

sprite.h
```diff
@@ -29,20 +29,41 @@ class Sprite : public Identifiable<Empty> {
 	}
 
 	Point &home();
+	inline static bool allow_screen_wrapping = true;
```
pierce-smith1: No. No behavior (ESPECIALLY control flow) dictated by global variables. A static variable should pretty much always be constant. 

Now that we have to decide whether or not the screen wrapping takes place on a pattern-by-pattern basis, this means the responsibility for wrapping or clamping the sprite has now fallen to the sprite coordinator, NOT the sprite itself. I think it goes somewhere after `m_current_player->update()` in `SpriteChoreographer::update()`. This means the `wrap` and `keep_in_bounds` functions will probably move over to that class as well.

spritecontrol.cpp
```diff
@@ -183,17 +185,19 @@ std::map<PatternName, SinglePassPlayer::MoveFunction> SinglePassPlayer::move_fun
 		double lateral_modifier = (directions[sprite->id()] & West) ? -1.0 : 1.0;
 		double vertical_modifier = (directions[sprite->id()] & South) ? -1.0 : 1.0;
 
-		get<X>(sprite->home()) += (offset / cfg[Cfg::TimeDivisor]) * lateral_modifier;
+		get<X>(sprite->home()) += offset / cfg[Cfg::TimeDivisor] * lateral_modifier;
```
pierce-smith1: Please try to keep meaningless changes to a minimum. It makes it harder to review because a really impactful change and a completely pointless one scream at me in the exact same way. Leave the style changes to a cleanup PR.

pierce-smith1 (changes requested):
I haven't tested it but this won't fly. I'm getting really worried about complexity in our codebase.

sprite.h
McAwesome123: I'll just quickly mention that my first idea was for it to check the current pattern, but couldn't figure out how to make it work without changing approximately 50 different things.

However, I was also not changing the location of the wrap function.

noise.cpp
pierce-smith1: Since this function is so highly specialized now, I don't think it belongs in Noise anymore, it probably should go in SpriteWiggler since that's what it has been redesigned for.

sprite.h
```diff
@@ -45,6 +46,33 @@ class Sprite : public Identifiable<Empty> {
...
 
+class SpriteWiggler {
+public:
+	SpriteWiggler();
+	void randomize_tendency();
+	void wiggle(Context &ctx, const Point &home, Point &relpos, const double &magnitude);
+
+private:
+	template<typename T>							// first value gets multiplied or divided by 1 + random(second value)
+	using pair_randomness = std::pair<T, double>;	// probably needs a better name :)
```
pierce-smith1: You might consider getting rid of this template alias entirely and just writing out `std::pair<double, size_t>` or whatever. It makes things a little simpler and you don't have to think of a better name :)

spritecontrol.cpp
```diff
@@ -119,6 +120,46 @@ void PatternPlayer::set_pattern(PatternName pattern) {
 	m_hash_offset++;
 }
 
+void PatternPlayer::handle_off_screen_sprites(bool wrap_sprites) {
```
pierce-smith1: Rather than having a single function that does two dramatically different things based on a boolean parameter, consider splitting this into two functions and then deciding which one to call based on what you would have passed to this function. 
```cpp
// SpriteChoreographer::update()
auto should_wrap = non_screen_wrapping_patterns.contains(m_pattern);
if (should_wrap) {
  m_current_player->wrap_off_screen_sprites();
} else {
  m_current_player->clamp_off_screen_sprites();
}
```
This spreads out responsibility more so that `handle_off_screen_sprites` isn't doing so many things. Plus, it always feels a little icky to have a function that takes a single boolean because it isn't clear from the call site what the difference between passing `true` and `false` would be or what that parameter actually means.

noise.cpp
```diff
@@ -124,14 +124,20 @@ double PerlinNoise::interpolate(double a, double b, double w) {
...
+	auto inverse_lerp = [](double a, double b, double t) -> double { return (t - a) / (b - a); };	// Interpolate from 0 to 1 as t ranges from a to b
+	int increase = (random() < 0.5 ? +1 : -1);
 
-	if (up) {
-		return base + random() * (max - base) * step;
+	double scaled_distance;
```
pierce-smith1: Consider immediately initializing this variable with a ternary...
```cpp
double scaled_distance = base < center 
    ? -(1 - pow(1 - inverse_lerp(center, min, base), distance_exp));
    : +(1 - pow(1 - inverse_lerp(center, max, base), distance_exp));
```
...so that `scaled_distance` never exists in an uninitialized state. We shouldn't give ourselves the opportunity to accidentally use an uninitialized variable.

noise.cpp
```diff
@@ -124,14 +124,20 @@ double PerlinNoise::interpolate(double a, double b, double w) {
...
+	if (base < center) {
+		scaled_distance = -(1 - pow(1 - inverse_lerp(center, min, base), distance_exp));
 	} else {
-		return base - random() * (base - min) * step;
+		scaled_distance = +(1 - pow(1 - inverse_lerp(center, max, base), distance_exp));
 	}
+
+	return std::clamp(base + (1 - pow(1 - random(), rand_exp)) * (increase - scaled_distance) * step, min, max);
```
pierce-smith1: I would extract that `base + (1 - pow ...` expression to its own variable. It's a pretty complicated expression and having it in an intermediate variable will help with clarity and debugging.

sprite.cpp
```diff
@@ -9,7 +9,11 @@
 using std::get;
 
 Sprite::Sprite(const Texture *texture, const Point &home, const bool has_trail)
-	: m_texture(texture), m_home(home), m_relpos(0.0, 0.0), m_size(cfg[Cfg::SpriteSize] / 1000.0f), m_trail_start_index(0)
+	: m_texture(texture),
+	m_home(home),
+	m_relpos(0.0, 0.0),
+	m_size(get_size()),
```
pierce-smith1, line 15: I would be careful about calling a member function in an member initializer list - it's possible that the function is called before all the members have a chance to initialize, and if the function uses an uninitialized member you're in shitsville. That's not going to happen in this case because `get_size` doesn't use any members, but we open the door to make that mistake in the future, so consider inlining the function call here or making `get_size` static.

sprite.cpp
```diff
@@ -99,7 +91,84 @@ void Sprite::draw_trail(Context &ctx) {
...
-Yonker::Yonker(const Texture *texture, const Point &home) 
+SpriteWiggler::SpriteWiggler()
+	: m_relpos_tendency(0, 0),
+	m_frames_when_tendency_changes(0),
+	m_distance_from_tendency_bias(default_distance_from_tendency_bias.first),
+	m_tendency_distance_from_home_bias(default_tendency_distance_from_home_bias.first),
+	m_wiggle_amount_bias(default_wiggle_amount_bias.first)
+{ }
+
+void SpriteWiggler::wiggle(Context &ctx, const Point &home, Point &relpos, const double &magnitude) {
```
pierce-smith1: Taking small objects like a double by reference may be less efficient than just copying it.  The rule of thumb is to take classes by reference but not primitives.

pierce-smith1 (approved):
What a huge relief. The code is looking really solid now and I didn't find anything in testing.  Below is a bunch of really petty nitpicks you can choose to fix, or not.

spritecontrol.cpp
McAwesome123: nitpicking your nitpick: `should_wrap` should be the opposite here.

yeah, I get the bool parameter thing. the next best thing I can think of (while keeping it one function) is to use an enum instead, but that also feels rather unnecessary?

sprite.cpp
McAwesome123: I'm not inlining it to avoid having what's supposed to be the same equation being defined in two places.
Also, `get_size` is already static.

pierce-smith1: My mistake, it's fine then

PR #73, opened 2024-06-19
Alternative yoker trails 

McAwesome123:
This PR adds an alternative trail type - one that loosely follows the parent sprite rather than copying its exact position.


McAwesome123:
To temper your expectations somewhat:
The current code is at the stage of "working" and I'd not describe it as being "good" in any reasonable capacity.

pierce-smith1:
> To temper your expectations somewhat:
The current code is at the stage of "working" and I'd not describe it as being "good" in any reasonable capacity.

Given this I'm not going to give a review or test it quite yet.

McAwesome123:
I'm unsure how much it'll help exactly, but it is currently very tempting to rebase this onto #72

McAwesome123:
> I'm unsure how much it'll help exactly, but it is currently very tempting to rebase this onto #72

It did not help.

Either way, I'm asking for your input as I feel like if I keep going, I'd just be banging my head into a brick wall.

sprite.cpp
```diff
@@ -67,13 +67,96 @@ void Sprite::transform() {
...
 
-	get_trail() = TrailSprite(m_texture, Point(final<X>(), final<Y>()));
-	increment_trail_index();
+	if (cfg[Cfg::TrailsExactFollow] == 1.0) {
+		get_trail() = TrailSprite(m_texture, Point(final<X>(), final<Y>()));
+		increment_trail_index();
+	} else {
+		// :cvheadache:
+
+		const double edge_boundary = 0.15 + m_size / 1.1;
```
pierce-smith1, line 81: All of these constants should be moved to a single place since they're used elsewhere too, namely in `PatternPlayer::wrap_off_screen_sprites` and its brother. `PatternPlayer` should probably just expose these somehow.

sprite.cpp
```diff
@@ -67,13 +67,96 @@ void Sprite::transform() {
...
+
+		auto get_target = [=](Point current_target, Point last_target) -> Point {
+			Point new_target = last_target;
+
+			get<X>(current_target) += horizontal_boundary;
+			get<Y>(current_target) += vertical_boundary;
+			get<X>(last_target) += horizontal_boundary;
+			get<Y>(last_target) += vertical_boundary;
+			for (; get<X>(last_target) < 0.0; get<X>(last_target) += horizontal_boundary * 2);
+			for (; get<Y>(last_target) < 0.0; get<Y>(last_target) += horizontal_boundary * 2);
```
pierce-smith1, line 95: As a matter of style I really don't like bodyless for loops. The "signature" is supposed to describe what is being looped over and the body is supposed to describe what the loop does. When you violate that convention I have to start using my brain to read the code which is bad.

sprite.cpp
```diff
@@ -67,13 +67,96 @@ void Sprite::transform() {
...
+			} else if (abs((get<Y>(current_target) - horizontal_boundary * 2) - get<Y>(last_target)) < abs((get<Y>(current_target) + horizontal_boundary * 2) - get<Y>(last_target))) {
+				get<Y>(new_target) += (get<Y>(current_target) - horizontal_boundary * 2) - get<Y>(last_target);
+			} else {
+				get<Y>(new_target) += (get<Y>(current_target) + horizontal_boundary * 2) - get<Y>(last_target);
+			}
+
+			return new_target;
+		};
+
+		auto wrap = [](double position, double min, double max) -> double {
```
pierce-smith1, line 119: Like the `edge_boundary` and `*_correction` constants, we should probably just have a single `wrap` function somewhere at this point.

sprite.h
```diff
@@ -110,8 +110,12 @@ class TrailSprite : public Sprite {
 
 	virtual void update(Context &ctx) override;
 
-	static int get_trail_length();
-	static int get_trail_space();
+	static size_t get_trail_length();
+	static double get_trail_space();
+
+	const Texture *get_texture();
```
pierce-smith1, line 116: This method doesn't seem to be used.

sprite.h
```diff
@@ -110,8 +110,12 @@ class TrailSprite : public Sprite {
...
+
+	Point target_position;
```
pierce-smith1, line 118: The trail function can be re-written in a way that doesn't need this. It's probably best to keep this TrailSprite class as simple as possible because it is meant to be very dumb.

sprite.cpp
```diff
@@ -67,13 +67,96 @@ void Sprite::transform() {
...
+			trail.target_position = get_target(Point(final<X>(), final<Y>()), trail.target_position);
+			move_trail(trail.home(), trail.target_position);
+			trail.change_texture(m_texture);
+		}
+		for (size_t i = TrailSprite::get_trail_length() - 1; i > 0; i--) {
+			TrailSprite &trail = get_trail(i - 1);
+			const TrailSprite &target_trail = get_trail(i);
+			trail.target_position = get_target(Point(target_trail.final<X>(), target_trail.final<Y>()), trail.target_position);
+			move_trail(trail.home(), trail.target_position);
+			trail.change_texture(m_texture);
```
pierce-smith1, line 157: I don't know exactly what this code does because I seem to be incapable of reading it, but it seems unnecessarily complicated and it has some weird behavior when sprites wrap. I tried to re-think the problem myself to see if I could discover what is going on here and if it can be improved and I think I found some potential improvements.

To get this smooth movement effect we're after, all we need to do is move each sprite in the trail a proportion of the distance between it and the next one, then make the last sprite in the trail is always on top of the main sprite. That's literally it. The actual hard part (as I'm sure you discovered) is dealing with the wrapping.

The way I thought of it was like this: when a trail sprite wraps, imagine there's a "ghost" sprite that does not wrap and keeps going off-screen. We want the previous trail sprite to follow that ghost instead of the actual one, since it's all the way at the other side of the screen now. Thankfully, the position of the ghost is pretty easy to calculate - it's just the position of the wrapped sprite moved (up/down/left/right) by one entire screen. 

So, one thing we could do is compute four ghosts of the next sprite in each +/- x/y directions, see if any of those ghosts are _closer_ than the sprite itself, and if they are, follow them instead. If the sprite starts following a ghost, eventually it will go offscreen and get wrapped itself, at which point it will start following the real sprite again and all is well. 

I then realized that instead of computing four ghost locations, it is equivalent to just ask if the sprite is more than half of a full screen width or height away, and if so, pull it back by a screen width/height to find the ghost we should be following. 

I implemented this over here and it seems to work pretty seamlessly: https://github.com/McAwesome123/yok.scr/pull/13

sprite.cpp
```diff
@@ -67,13 +67,96 @@ void Sprite::transform() {
...
+			Point new_target = last_target;
+
+			get<X>(current_target) += horizontal_boundary;
+			get<Y>(current_target) += vertical_boundary;
+			get<X>(last_target) += horizontal_boundary;
+			get<Y>(last_target) += vertical_boundary;
+			for (; get<X>(last_target) < 0.0; get<X>(last_target) += horizontal_boundary * 2);
+			for (; get<Y>(last_target) < 0.0; get<Y>(last_target) += horizontal_boundary * 2);
+			get<X>(last_target) = fmod(get<X>(last_target), horizontal_boundary * 2);
+			get<Y>(last_target) = fmod(get<Y>(last_target), vertical_boundary * 2);
```
pierce-smith1, line 97: So as I mentioned I have like 0 idea what any of this code is doing and I can't figure it out. Why are we `fmod`ing here? What's the huge cluster of if statements for? I think I just need a short explanation of the high level idea here and then I'll get it.

pierce-smith1 (commented):
First some design thoughts and then some code thoughts.

On design: I'm not sure this option really justifies its own existence. For almost all patterns, the difference between this and exact trails is almost indistinguishable. It shines in Bubbles and probably Boids/Rivers, but in most other patterns yokins are moving in wide, slow arcs, which make the trails look very similar. In fact for the Bouncy pattern, it actually ends up looking really strange. It also goes against the classic inspiration for trails, so it's sort of inelegant and confusing. 

The cost of this is increased UI and code complexity, and I'm not certain either of those are worth it. I am especially concerned about death from a thousand cuts in our UI. If this change makes it in, maybe it should not be an option - maybe we should just choose which kind of trails we like better and leave it at that.

Now of course this is just a silly screensaver so I am willing to put all of that aside and YOLO it into the codebase like a lot of other features, but every time I do so the future me that is making yoksdotnet gets a little grumpier that I have yet another feature to port over.

Code comments below.

sprite.cpp
McAwesome123: I uh
think that's basically how I was approaching this
though I'd calculate the ghost sprite's positions and use the closest one or something
The issue I had is that it would break if the trail sprite ended up lagging so far behind that the actual sprite would be closer than the ghost sprite and it'd end up suddenly changing directions

sprite.cpp
McAwesome123: I _think_ what I'm trying to do there is figure out the spot the trail is supposed to be chasing, while accounting for the screen wrapping.
First, I try to get the numbers to be between 0 and 2 * boundary, using `for` if above and `fmod` if below.
Then I think I try to figure out which point is closest to the sprite's current position; either the last target position as it is now, after being wrapped one way, or after being wrapped the other way.
And the difference between the current position and the closest point is added to what the last target position was and that's the new target position.

PR #74, opened 2024-06-19
Adding a Pattern Selector

McAwesome123:
This PR adds the option to enable and disable patterns, as well as being able to select a random starting pattern.
Also, starting emotions are now random.


McAwesome123:
For now, I've sort of just stuffed all of the pattern config into SpriteChoreographer, but there could very well be a much better place to put it.

It's also very messy, as you might expect.

spritecontrol.cpp
```diff
@@ -90,8 +94,87 @@ void SpriteChoreographer::update() {
...
 
+std::vector<std::pair<PatternName, std::wstring>> SpriteChoreographer::get_disabled_patterns() {
+	return m_disabled_patterns;
+}
+
+std::vector<std::pair<PatternName, std::wstring>> SpriteChoreographer::get_enabled_patterns() {
+	return m_enabled_patterns;
+}
+
+void SpriteChoreographer::load_disabled_patterns() {
```
pierce-smith1: You're right, `SpriteChoreographer` is not the right place for this. Perhaps there should be a `PatternRepository` or similar whose sole responsibility is to save and load patterns from the registry - similar to the `PaletteRepository`, but quite a bit simpler. 

I also think it would be better to not store an `m_disabled_pattern` list as a member and "load" stuff from the registry by putting it into that member. The PatternRepository should do two really simple jobs: 1), read the registry into a list and return it, and 2), save a list to the registry. No state management is necessary here - it should be the responsibility of the code that uses the PatternRepository to keep track of any state it needs. For example in `ScreenSaverPatternSelectorDialog` you have those static vectors that temporarily hold the pattern data before it is saved - do things like that instead of having an `m_disabled/enabled_patterns` member and a `change_disabled_patterns` function.

spritecontrol.cpp
```diff
@@ -90,8 +94,87 @@ void SpriteChoreographer::update() {
 	}
 }
 
+std::vector<std::pair<PatternName, std::wstring>> SpriteChoreographer::get_disabled_patterns() {
+	return m_disabled_patterns;
+}
+
+std::vector<std::pair<PatternName, std::wstring>> SpriteChoreographer::get_enabled_patterns() {
```
pierce-smith1: Why are we sending around these pairs instead of just the `PatternName`s by themselves? If the calling code needs the string names of the patterns, it can just look them up in the `pattern_names` map.

If we do need to do this for some reason, we should probably turn that std::pair into a class so the two pieces of data can get more meaningful names, especially since this is a public interface.

spritecontrol.cpp
```diff
@@ -90,8 +94,87 @@ void SpriteChoreographer::update() {
...
+	return m_enabled_patterns;
+}
+
+void SpriteChoreographer::load_disabled_patterns() {
+	const std::wstring disabled_patterns_name = L"DisabledPatterns";
+	const std::wstring disabled_patterns_default = L"";
+	const std::wstring disabled_patterns_string_delimiter = L",";
```
pierce-smith1: Reading lists from the registry is starting to become a common concern for us. I wonder if we should have a function on the Registry type that saves and loads lists of strings, since the act of splitting and joining on delimiters and shit is adding a lot of noise to this code here and probably other places too.

configdialog.cpp
```diff
@@ -1530,3 +1547,128 @@ LRESULT CALLBACK AddPredefinedPaletteDialog(HWND dialog, UINT message, WPARAM wp
 
 	return FALSE;
 }
+
+LRESULT CALLBACK ScreenSaverPatternSelectorDialog(HWND dialog, UINT message, WPARAM wparam, LPARAM lparam) {
+	static std::vector<std::pair<PatternName, std::wstring>> disabled_patterns = { };
```
pierce-smith1, line 1552: I think I started this trend, but looking at it now storing important state as statics in a window procedure function is starting to feel kind of icky. If we ever have two windows that use these proc functions they will each be reading from the same data and that might be wrong. Maybe. Something to think about later.

noise.cpp
```diff
@@ -29,12 +30,15 @@ PerlinNoise::Vector PerlinNoise::Vector::sub(const Vector &v) const {
...
+// std::chrono is a fucking pain
+const unsigned int PerlinNoise::base_seed = cast<unsigned int>(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count());
+
 PerlinNoise::Vector PerlinNoise::grad_vector(double x, double y, double z) {
 	// To grow a seed from a single point
 	// With orderly chaos a constant we anoint
 	// But be careful not to overflow
 	// For what happens then, well, no one knows
-	double seed = 1;
+	double seed = base_seed;
```
pierce-smith1, line 41: Honestly I'm surprised the screensaver still works after doing this. I'm assuming this is to solve the perlin noise pattern being the same every time?

It would be simpler to just make `base_seed` a static local instead of a static member, I don't think there is any reason it has to be a static member.

pierce-smith1 (changes requested):
This feature is AWESOME, I love it. The implementation is good too, you thought of a lot of small details like turning off the change interval slider if there's only 1 pattern. I think there is some work to do on the code but I really like where it's headed.

configdialog.cpp
```diff
@@ -1530,3 +1547,128 @@ LRESULT CALLBACK AddPredefinedPaletteDialog(HWND dialog, UINT message, WPARAM wp
...
+	static std::vector<std::pair<PatternName, std::wstring>> enabled_patterns = { };
```
pierce-smith1: I'm not sure it makes much sense to have both a disabled patterns list AND an enabled patterns list. From what I can tell, this is forcing you to do a lot of extra bookkeeping to make sure these two lists stay in sync - this is a problem because it's complicated and noisy and what happens if you get it wrong? All hell will surely break loose if a pattern shows up in _both or neither_ the disabled set or the enabled set,

The set of enabled patterns can be easily computed from the set of disabled patterns - it's just the set of all patterns minus the set of disabled patterns. So, whenever someone needs to know the enabled patterns, I think they should just compute that themselves rather than relying on this code to keep track of it. 

(As for the Random pattern, anything that deals with saving/loading patterns should probably just pretend that Random doesn't exist. As long as it doesn't get put into the dialog, it should be fine. This frustrates the hell out of me because if C++ had an actual useful type system like Typescript, we could easily model the fact that the set of all pattern names is a strict subset of the set of all choices the user can make for starting pattern, but in C++ land that would require another useless enum and horrible abuses of `std::variant` that are not worth it.)

spritecontrol.cpp
McAwesome123: Would probably be a good idea.

spritecontrol.cpp
McAwesome123: I'm... not sure. Either it didn't work or (more likely) I didn't realize that.

spritecontrol.h
```diff
@@ -89,10 +90,26 @@ class SpriteChoreographer {
 	bool should_change_pattern();
 	void update_player();
 
+	inline static std::vector<PatternName> m_enabled_patterns = { };
```
pierce-smith1: It there is reason this is static? It probably shouldn't be especially since it's being set in the constructor.

spritecontrol.cpp
```diff
@@ -80,6 +80,10 @@ SpriteChoreographer::SpriteChoreographer(PatternName choreography, Sprites *spri
 	: m_pattern(choreography), m_ctx(ctx), m_sprites(sprites)
 { 
 	m_players = { new SinglePassPlayer(sprites, ctx), new GlobalPlayer(sprites, ctx) };
+	m_enabled_patterns = PatternRepository::load_enabled_patterns();
```
pierce-smith1: Passing a pattern into the constructor only for it to possibly ignore it and give you a different one is poor form. If I asked for X, I should get X. There's no way we can guarantee this through the type system but I think we can just assume a pattern being passed in is in the enabled patterns list, since that helps expose bugs happening elsewhere instead of being all coy about it.

It's also worth considering whether or not m_enabled_patterns should be set by a parameter - I'm not really sure whose responsibility it is to set this. Either option is probably fine.

Resource.rc
```diff
@@ -117,6 +116,8 @@ BEGIN
     PUSHBUTTON      "Customize",IDC_PALETTE_CUSTOMIZE,292,79,50,14
     CONTROL         "Use custom palette set",IDC_CUSTOM_PALETTES_CHECK,
                     "Button",BS_AUTOCHECKBOX | WS_TABSTOP,293,64,89,10
+    PUSHBUTTON      "Open",IDC_PATTERN_SELECTOR_OPEN,336,132,50,14
```
pierce-smith1, line 119: I think we can do something similar to the palette customization here and just label the button "Customize" and get rid of the "Palette Selector" label. It's better to focus on what the button does rather than what it is.

spritecontrol.cpp
```diff
@@ -332,3 +337,50 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
 		}
 	}},
 };
+
+std::vector<PatternName> PatternRepository::load_disabled_patterns() {
+	Registry registry;
+
+	std::vector<PatternName> disabled_patterns = { };
+
+	std::vector<std::wstring> disabled_patterns_strings = split<std::wstring>(registry.get_string(disabled_patterns_name, disabled_patterns_default), disabled_patterns_string_delimiter);
```
pierce-smith1, line 350: This shit is so noisy and hard to read but it's just C++ being C++, I have no suggestions for doing it differently and would have done it the same :(

spritecontrol.cpp
```diff
@@ -101,7 +105,8 @@ bool SpriteChoreographer::should_change_pattern() {
 }
 
 void SpriteChoreographer::change_pattern() {
-	m_pattern = (PatternName) (Noise::random() * cast<double>(_PATTERN_COUNT));
+	size_t rand = (size_t) (Noise::random() * m_enabled_patterns.size());
+	m_pattern = m_enabled_patterns.at(rand);
```
pierce-smith1: I think we should make it so that the choreographer can never change into the same pattern it's already playing, since that behavior feels pretty unintuitive. If I ask for two patterns I would expect it to just swap between those patterns, for example.

spritecontrol.h
```diff
@@ -89,10 +90,26 @@ class SpriteChoreographer {
...
 	PatternName m_pattern;
 	std::vector<PatternPlayer *> m_players;
 	PatternPlayer *m_current_player;
 };
 
+class PatternRepository {
+public:
+	static std::vector<PatternName> load_disabled_patterns();
+	static std::vector<PatternName> get_enabled_patterns(const std::vector<PatternName> &disabled_patterns);
+	inline static std::vector<PatternName> load_enabled_patterns() { return get_enabled_patterns(load_disabled_patterns()); }
```
pierce-smith1: Even though it's a really simple function I would move this to a .cpp file like the rest of them. It just makes it easier to look for and find, I got lost since I didn't realize you'd defined it inline here.

configdialog.cpp
```diff
@@ -1530,3 +1557,128 @@ LRESULT CALLBACK AddPredefinedPaletteDialog(HWND dialog, UINT message, WPARAM wp
...
+					if (enabled_patterns.size() > 0) {
+						ShowWindow(no_patterns_enabled_text, SW_HIDE);
+						EnableWindow(ok_button, TRUE);
+					}
+
+					return FALSE;
+				}
+				case IDC_DISABLE_PATTERN: {
+					while (ListBox_GetSelCount(enabled_patterns_listbox) > 0) {
+						int selection = 0;
```
pierce-smith1, line 1662: Add some line breaks in there man, it makes a huge difference for readability. Here's how I would space things out (trying to keep the same "steps" of the operation visually together)

```cpp
int selection = 0;
ListBox_GetSelItems(disabled_patterns_listbox, 1, &selection);

auto &selected_pattern = disabled_patterns.at(selection);
enabled_patterns.push_back(selected_pattern);
ListBox_AddString(enabled_patterns_listbox, pattern_strings.at(selected_pattern).c_str());

ListBox_DeleteString(disabled_patterns_listbox, selection);
disabled_patterns.erase(std::find(disabled_patterns.begin(), disabled_patterns.end(), selected_pattern));
```

spritecontrol.cpp
pierce-smith1: > If I asked for X, I should get X

Case in point... I just spent at least 15 minutes looking for where you were handling the random pattern case in here, and it turns out it's just _super_ subtly handled by this. Don't be so sneaky, handle that case explicitly!

pierce-smith1 (changes requested):
Again, I love it, this feature is awesome. I didn't find anything broken in testing or combing over the code itself, so very well done. 

I would like a few more things to get addressed though before I feel comfortable approving.

spritecontrol.h
McAwesome123: The reason is _probably_ that it used to be static before I moved all the config stuff to its own class and I just didn't bother to change it.

Resource.rc
McAwesome123: I sorta feel like just having a "customize" button would give the impression that you can actually customize pattern parameters, which is mainly why I went with this instead.

spritecontrol.cpp
McAwesome123: Yeah, probably. As far as I'm aware changing into the same pattern was always possible, but it's definitely a lot more noticeable with 2 patterns than with 8.

Resource.rc
pierce-smith1: That's a valid concern - maybe this is fine then cause I really can't think of anything better.

spritecontrol.cpp
```diff
@@ -105,8 +104,13 @@ bool SpriteChoreographer::should_change_pattern() {
 }
 
 void SpriteChoreographer::change_pattern() {
-	size_t rand = (size_t) (Noise::random() * m_enabled_patterns.size());
-	m_pattern = m_enabled_patterns.at(rand);
+	std::vector<PatternName> patterns = { };
```
pierce-smith1: A better name for this is probably `candidate_patterns` or similar, just calling it `patterns` is really undescriptive.

spritecontrol.cpp
```diff
@@ -105,8 +104,13 @@ bool SpriteChoreographer::should_change_pattern() {
...
+	std::copy_if(m_enabled_patterns.begin(), m_enabled_patterns.end(), std::back_inserter(patterns), [&](const PatternName &pattern) {
+		return pattern != m_pattern;
+	});
+
+	size_t rand = (size_t) (Noise::random() * patterns.size());
```
pierce-smith1: Don't know why I didn't notice this at first but this probably shouldn't be called `rand` since it's undescriptive and it conflates with std::rand - `random_index` might be better.

pierce-smith1 (approved):
Works for me.

PR #75, opened 2024-09-18
Aboiding

McAwesome123:
This PR adds a new "Boids" pattern.

The yokies will, for the most part, bunch together and move as a large group. Or occasionally, in a few smaller groups.


McAwesome123:
This will probably need a fair bit of clean-up and/or optimization.
I'm also not 100% sure everything functions as intended.

graphics.cpp
```diff
@@ -72,3 +72,59 @@ GLubyte *Texture::data() const {
 
 	return texture_data;
 }
+
```
pierce-smith1, line 75: I try to avoid operator overloading as much as possible, since it obscures function calls and this makes telling what is _actually_ going on in some code much harder.

A vector type like this is probably one of the most sensible uses but I still dislike it.

graphics.cpp
```diff
@@ -72,3 +72,59 @@ GLubyte *Texture::data() const {
...
+Point operator+(const Point &l, const Point &r) {
+	double x = get<X>(l) + get<X>(r);
+	double y = get<Y>(l) + get<Y>(r);
+
+	return Point(x, y);
+}
+
+Point operator+=(Point &l, const Point &r) {
```
pierce-smith1: An actual nitpick, if these operators stick around: assignment operators should return a reference. It eliminates the potential for a useless copy and is just generally the idiomatic way to do it.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+
+				double vision_radius = random_curve(0.5) * VISION_X_RADIUS;
+				vision_range[sprite->id()] = vision_radius;
+
+				double degrees = random_curve(1.0, 10.0, 0.25) * BLIND_DEGREES;
+				blind_angle[sprite->id()] = degrees * M_PI / 180.0;
+			}
+		}
+
+		std::map<Sprite *, Point> separation_velocity_changes;
```
pierce-smith1: These maps are probably hurting the performance the most. I don't have any super specific optimization plan but I think these eventually turn into static arrays, OR maybe disappear altogether (can we just apply the transformations now instead of storing them and applying them later?)

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+		std::map<Sprite *, Point> cohesion_velocity_changes;
+		std::map<Sprite *, Point> edge_push_velocity_changes;
+
+		for (Sprite *current_sprite : *sprites) {
+			Point current_final = Point(current_sprite->final<X>(), current_sprite->final<Y>());
+			size_t sprites_seen = 1;
+			double average_angle = get_angle(velocity[current_sprite->id()]);
+			Point average_pos = Point(current_final);
+
+			for (Sprite *other_sprite : *sprites) {		// dejil... i am sorry...
```
pierce-smith1: Im not sure how we can avoid a nested for loop, but we should think about it, since getting rid of it would be a massive performance win.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
 			glEnd();
 		}
 	}},
+	{ Boids, [](Sprites *sprites, Context *ctx, std::function<double(Id)> get_offset) {
+		const static double SCREEN_SIZE = (double) ((long long) ctx->rect().bottom * ctx->rect().right);
+		const static double STRETCH_RATIO = (double) (ctx->rect().bottom) / ctx->rect().right;
+		const static double SEPARATION_Y_RADIUS = (6.0 / (cfg[Cfg::SpriteCount] / 1.5 + 40.0)) * std::pow(SCREEN_SIZE / (1080LL * 1920LL) / 3.0 + 0.7, 1.1);
+		const static double SEPARATION_X_RADIUS = SEPARATION_Y_RADIUS * STRETCH_RATIO;
+		const static double VISION_Y_RADIUS = (10.0 / (cfg[Cfg::SpriteCount] / 15.0 + 20.0)) * std::pow(SCREEN_SIZE / (1080LL * 1920LL) / 3.0 + 0.7, 1.1);
```
pierce-smith1: Did I really hard code this shit for 1920x1080? what a disaster? I truly did not give a shit.

pierce-smith1 (commented):
Pretty much everything that really sucks about this is inherited from bubbles. I don't have much to say about the math right now since if math works then it works. But before this goes in let's try to clean up some of the mess bubbles made.

Also i read this code on my phone on a screen that can barely fit one word of text so there are probably nitpicks im missing.

graphics.cpp
McAwesome123: the main reason I added these is cause adding vectors together and multiplying them by a constant make sense to me and I just don't wanna write `get<X>(...) += ...; get<Y>(...) += ...;` every single time.

spritecontrol.cpp
McAwesome123: I imagine these could probably be static arrays that's the same size as the sprites vector.
The main reason I'm not applying the transformations here is cause that'd mean later sprites read different values than earlier sprites. but whether that's actually an issue I'm not sure.

spritecontrol.cpp
McAwesome123: It'd help a lot but I have no clue how to actually get rid of it.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+		auto random_curve = [](double variation, double exponent = 5.0, double slope = 0.1) {
+			double random = Noise::random();
+			double sign = Noise::random() < 0.5 ? -1.0 : 1.0;
+			random = random * slope + pow(random, exponent) * (1 - slope);
+			return 1.0 + sign * random * variation;		// Equation: 1 + sign(rand) * (|rand| * slope + |rand|^exp * (1 - slope)) * variation
+		};
+
+		// Converts a vector to an angle in the range (-M_PI, M_PI].
+		auto get_angle = [](const Point &vector) {
+			double x = get<X>(vector);
```
pierce-smith1: Neat trick from C++17 I haven't used at all, but I use the corresponding feature in javascript all the time - you can do this:
```cpp
auto [x, y] = vector;
// or, better if you don't mutate the variables (as in this function)
const auto &[x, y] = vector;
```
This is called structured binding and it's a benefit we get from Point just being a std::pair. Cool!

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+			double y = get<Y>(vector);
+
+			if (x == 0 && y == 0) {
+				return 0.0;
+			}
+
+			double angle = atan2(y, x);
+
+			// i'm sure there will be absolutely nothing wrong with this
+			if (x >= 0) {
```
pierce-smith1: Why these if checks and -+pi adjustments?

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+			} else if (y >= 0) {
+				return angle + M_PI;
+			} else {
+				return angle - M_PI;
+			}
+		};
+
+		// Wraps angles to be inside the range (-M_PI, M_PI]
+		auto wrap_angle = [](double angle) {
+			angle += M_PI;
```
pierce-smith1: This is some wack ass logic. Why not just
```c++
if (angle > pi) return angle - 2pi;
if (angle < -pi) return angle + 2pi;
return angle;
```
This is only half a rhetorical question because I'm not 100% sure the above works, but I feel like it should..?

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+		};
+
+		auto get_vector_magnitude = [](const Point &vector) {
+			double x = get<X>(vector);
+			double y = get<Y>(vector);
+
+			return sqrt(x * x + y * y);
+		};
+
+		auto normalize_vector = [get_vector_magnitude](const Point &vector) {
```
pierce-smith1: Nitpick, when capturing from a lambda it pays to just use `[&]` instead of writing everything out explicitly. Plus, something like another lambda really should be captured by reference.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+
+		const static double DEFAULT_FORCE_MULT = 0.05;	// Multiplier for all forces below
+		const static double SEPARATION_FORCE_MULT = DEFAULT_FORCE_MULT * 1.5;			// How strongly to separate sprites that are too close
+		const static double ALIGNMENT_FORCE_MULT = DEFAULT_FORCE_MULT * 1.2;			// How strongly to align sprites that are in a pack
+		const static double COHESION_FORCE_MULT = DEFAULT_FORCE_MULT * 0.5;				// How strongly to pull sprites towards the middle of their pack
+		const static double EDGE_PUSH_FORCE_MULT = DEFAULT_FORCE_MULT * 1.5;			// How strongly to push sprites away from the edges
+		const static double DESIRED_VELOCITY_RETURN_MULT = DEFAULT_FORCE_MULT * 0.3;	// How strongly to accelerate sprites towards their desired velocity
+
+		static std::map<Id, Point> velocity;
+		static std::map<Id, double> desired_velocity;
```
pierce-smith1: Nitpick, this should probably be called something like `desired_speed` to emphasize it is not a "velocity" in the sense that is used elsewhere in the code.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+			Point average_pos = Point(current_final);
+
+			for (Sprite *other_sprite : *sprites) {		// dejil... i am sorry...
+				if (current_sprite == other_sprite) {
+					continue;
+				}
+
+				Point other_final = Point(other_sprite->final<X>(), other_sprite->final<Y>());
+				Point diff = current_final - other_final;
+				get<Y>(diff) *= STRETCH_RATIO;
```
pierce-smith1: This STRETCH_RATIO shit is really dumb. What we should have is a coordinate system that actually makes sense rather than the stupid -1 to 1 but-its-actually-stretched-out-into-a-rectangle system we are using now. I guess we can just continue doing it for now but let's hope to remove it in the future.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+
+		for (auto &sprite : alignment_velocity_changes) {
+			Point &sprite_velocity = velocity[sprite.first->id()];
+			double magnitude = get_vector_magnitude(sprite_velocity);
+			sprite_velocity /= magnitude;
+
+			Point diff = sprite.second - sprite_velocity;
+			diff *= ALIGNMENT_FORCE_MULT;
+
+			sprite_velocity = normalize_vector(sprite_velocity + diff) * magnitude;
```
pierce-smith1, line 553: This looks like an innocent variable assignment, but actually it modifies a collection! I would avoid doing stuff like this and just write out `velocity[sprite.first->id()] = whatever` explicitly. That is, if these collections even continue existing, which they probably won't.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+				if (abs(relative_angle) >= blind_angle[current_sprite->id()]) {
+					sprites_seen++;
+					double other_angle = get_angle(velocity[other_sprite->id()]);
+					average_angle += other_angle;
+					average_pos += other_final;
+				}
+
+				if (dist < desired_separation[current_sprite->id()]) {
+					diff /= dist * dist / desired_separation[current_sprite->id()];
+					separation_velocity_changes[current_sprite] = diff;
```
pierce-smith1: Thoughts on optimization; what if instead of storing these changes in a map and applying them later, we just applied them right now? Same for the other adjustments being stored in maps. It would have a butterfly effect on the rest of the computations, but it probably won't be that bad because they are all very subtle changes.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+		const static double EDGE_PUSH_X_RADIUS = EDGE_PUSH_Y_RADIUS * STRETCH_RATIO;
+
+		const static double DEFAULT_FORCE_MULT = 0.05;	// Multiplier for all forces below
+		const static double SEPARATION_FORCE_MULT = DEFAULT_FORCE_MULT * 1.5;			// How strongly to separate sprites that are too close
+		const static double ALIGNMENT_FORCE_MULT = DEFAULT_FORCE_MULT * 1.2;			// How strongly to align sprites that are in a pack
+		const static double COHESION_FORCE_MULT = DEFAULT_FORCE_MULT * 0.5;				// How strongly to pull sprites towards the middle of their pack
+		const static double EDGE_PUSH_FORCE_MULT = DEFAULT_FORCE_MULT * 1.5;			// How strongly to push sprites away from the edges
+		const static double DESIRED_VELOCITY_RETURN_MULT = DEFAULT_FORCE_MULT * 0.3;	// How strongly to accelerate sprites towards their desired velocity
+
+		static std::map<Id, Point> velocity;
```
pierce-smith1: Thoughts on optimization: we could avoid having these maps entirely if the velocity, desired separation, vision range, etc. were properties of the Sprites themselves. (Not literally on the `Sprite` type, since we don't want to pay the cost of this extra information everywhere when it's only used for this one function) That would require a decent architecture change, obviously, but I can envision some system that is able to transform regular Sprites into some custom type for each move function before it is called for the first time. That would actually go a long way for the other movement functions too! Probably out of scope for this PR.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+			sprite_velocity /= magnitude;
+
+			Point diff = sprite.second - sprite_velocity;
+			diff *= ALIGNMENT_FORCE_MULT;
+
+			sprite_velocity = normalize_vector(sprite_velocity + diff) * magnitude;
+			// there's probably a better way to do this but i can't be bothered figuring it out now
+		}
+
+		for (auto &sprite : cohesion_velocity_changes) {
```
pierce-smith1: More on structured binding: you can write the for loop like this:
```cpp
for (auto &[sprite, velocity_change] : alignment_velocity_changes) {
```
And then what used to be `sprite.first` becomes `sprite` and what used to be `sprite.second` becomes `velocity_change`. Basically, it gives you nicer variable names for free, which we should definitely use _if_ these for loops continue to exist.

spritecontrol.cpp
```diff
@@ -391,6 +392,247 @@ std::map<PatternName, GlobalPlayer::MoveFunction> GlobalPlayer::move_functions {
...
+			}
+
+			if (-1.0 + EDGE_PUSH_X_RADIUS > get<Y>(current_final)) {
+				get<Y>(edge_push) += ((-1.0 + EDGE_PUSH_X_RADIUS) - get<Y>(current_final)) / EDGE_PUSH_X_RADIUS;
+			} else if (1.0 - EDGE_PUSH_X_RADIUS < get<Y>(current_final)) {
+				get<Y>(edge_push) += ((1.0 - EDGE_PUSH_X_RADIUS) - get<Y>(current_final)) / EDGE_PUSH_X_RADIUS;
+			}
+
+			if (edge_push != Point(0.0, 0.0)) {
+				get<X>(edge_push) = pow(abs(get<X>(edge_push)), 1.0/2.5) * (get<X>(edge_push) < 0.0 ? -1.0 : 1.0);
```
pierce-smith1: Random thought: while having our Point type just be a std::pair is nice, I really have grown to hate the ergonomics of this `get<X>` crap, and really wish we could just do point.x and point.y instead. That is very possible but it's the scope of another PR.

pierce-smith1 (commented):
I read the code for real and I think I get what's going on. A bunch more scattered thoughts and nitpicks.

spritecontrol.cpp
McAwesome123: there is the possibility that I don't know how atan2 works

spritecontrol.cpp
McAwesome123: :lksix:

graphics.cpp
pierce-smith1: Oh yeah totally, you can still have convenience functions that hide all of that mess. I'm just saying, I would rather have:
```cpp
void add(Point &l, const Point &r);
Point sum(const Point &l, const Point &r);
```
instead of...
```cpp
Point &operator +=(Point &l, const Point &r);
Point operator +(Point &l, const Point &r);
```
I will admit that with the non-operator approach, the actual operation is a little less obvious because of the naming (it's not immediately clear if `add` or `sum` is the mutating operation, but it IS immediately clear which one of `+=` and `+` is the mutating operation. Still, that is why we have types.) But I still prefer it.

spritecontrol.cpp
pierce-smith1: > static arrays that's the same size as the sprites vector.

That's tough because the sprites vector is not static, we don't know how big it is until runtime. So we'd have to overcommit to a static array that is big enough to hold what we think the upper bound amount of sprites is, which is shitty from a lot of different perspectives.

spritecontrol.cpp
McAwesome123: technically there'd be the issue of "if the angle is <-3pi or >3pi then it'd wouldn't be in the proper range" but I'm not sure if that'd occur with the current usage

spritecontrol.cpp
McAwesome123: what exactly does that mean?

spritecontrol.cpp
pierce-smith1: Change the ifs into whiles then ;)

spritecontrol.cpp
McAwesome123: I absolutely agree.
And additionally, the fact that it goes from -1 to 1 instead of 0 to 1 massively fucked with me earlier cause I couldn't figure out why the fuck sprites on the bottom half and the left half of the screen were getting absolutely launched.

spritecontrol.cpp
pierce-smith1:
```cpp
std::vector numbers{1, 2, 3};
int &x = numbers[1];
x = 9; // Surprise! This looks like an innocuous assignment to a local variable, 
       // but actually it changes the value of numbers[1].
numbers[1] = 9; // Now this is much more obvious.
```
Generally surprises are a bad thing. It is surprising when an assignment changes something that is not even on the left hand side of the expression. This is where stupid bugs that eat an hour of your life come from.
As a general rule of thumb, you should avoid having a non-const reference as a local variable.

graphics.cpp
pierce-smith1: Hm, one thing I didn't immediately think about is because `Point` is not a real class, any functions we write for it have to be free, like the operators you made here. A free function that's just called `add` is really dumb. A free overload for the + operator is significantly less dumb.

Given that, here's what I'll say. I want to make some changes to the point class, which might come through in another PR, so for now, let's just leave things as you've made them here.

spritecontrol.cpp
McAwesome123: ah, right, array size needs to be known at compile time.
perhaps just a vector then, but I don't know.

spritecontrol.cpp
McAwesome123: I only read this now but that is what I've done

spritecontrol.cpp
McAwesome123: I mean, to be fair, modifying the collection is _exactly_ what I'm trying to do here, but that's fair.

McAwesome123:
Might still need some slight changes like adjusting parameters, etc, but I'd say it's close enough to being ready now.

pierce-smith1 (approved):
This is in really good shape now and I think it's appropriate to ship. There may be some minor issues with it but we can pick those up later if they come up.

PR #76, opened 2024-09-21
Make points not super annoying to use

pierce-smith1:
Sometimes simpler is better.


graphics.h
```diff
@@ -38,8 +38,7 @@ class Texture {
...
 
-using Point = std::pair<GLdouble, GLdouble>;
-enum Coord {
-	X = 0,
-	Y = 1
-};
\ No newline at end of file
+struct Point {
+	GLdouble x;
+	GLdouble y;
```
pierce-smith1, line 43: It turns out that structured binding is really smart and works on aggregates like this too, so we don't even need this to be a pair to get `auto [x, y] = point`.

pierce-smith1:
🚮

